PR #15, opened 2019-07-09
Add partners section

solazio:
## Done

- Add partners sections for home page

## Issue

Fixes https://github.com/canonical-web-and-design/web-squad/issues/1372


static/js/dynamic-image-load.js
```diff
@@ -0,0 +1,43 @@
+(function () {
+  var partners = document.querySelectorAll('.js-container');
```
anthonydillon: `js-container` seems vague. Can you use a more specific class such as `js-partner-logos` or something.

static/js/dynamic-image-load.js
```diff
@@ -0,0 +1,43 @@
...
+  var partnerLinks = document.querySelectorAll('.js-link');
```
anthonydillon: Same here, someone else could use these classes for a different purpose.

static/sass/_pattern_icons.scss
```diff
@@ -6,23 +6,23 @@ $color-icons-background: #999;
 @mixin canonical-p-icons {
   .is-dark {
     @include canonical-p-icon-facebook;
-    @include canonical-p-icon-twitter;
     @include canonical-p-icon-youtube;
+    @include canonical-p-icon-twitter;
   }
 }
 
 @mixin canonical-icon-facebook($background, $foreground) {
```
deadlight: You don't need to re-implement these mixins, you can use the ones from the core pattern, which now take foreground and background colours

deadlight:
Clicking on the logo filter links just takes the user to "#" and reverts to the top of the page.
Either handle or prevent the links' default behaviour.

PR #24, opened 2019-08-06
Add home page animations

### LINKED ISSUE #23: [c.c 19-32] Add animations to c.c home page
- [x] Add animations
- [x] Add links to the he logos in the `Product` section
- [x] Add more space after the last line of text in the footer

solazio:
## Done

- Add home page animations

## QA
- Go to [http://0.0.0.0:8002/](http://0.0.0.0:8002/) and make sure there are animations on the page.
- Please test it in multiple browsers.

## Issue

Fixes #23

### DIFF
--- a/static/sass/_pattern_animations.scss
+++ b/static/sass/_pattern_animations.scss
@@ -1,0 +1,55 @@
+@mixin canonical-u-animations {
+  .u-animation--slide-from-bottom {
+    opacity: 0;
+    animation: slide-from-bottom map-get($animation-duration, sleepy) linear
+      forwards;
+  }
+
+  .u-animation--slide-from-right {
+    opacity: 0;
+    animation: slide-from-right map-get($animation-duration, sleepy) linear
+      forwards;
+  }
+
+  .u-animation--slide-from-top {
+    opacity: 0;
+    animation: slide-from-top map-get($animation-duration, sleepy) linear
+      forwards;
+  }
+
+  @keyframes slide-from-bottom {
+    0% {
+      opacity: 0;
+      transform: translateY(2rem);
+    }
+
+    100% {
+      opacity: 1;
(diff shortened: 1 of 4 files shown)


static/sass/_pattern_footer.scss
```diff
@@ -13,6 +13,7 @@
     color: $color-mid-light;
     margin-top: -4rem;
     padding-top: 6rem;
+    padding-bottom: 2.5rem;
```
sowasred2012: This line should be above `padding-top`

static/sass/_pattern_animations.scss
```diff
@@ -0,0 +1,55 @@
+@mixin canonical-u-animations {
+  .u-animation--slide-from-bottom {
+    opacity: 0;
+    animation: slide-from-bottom map-get($animation-duration, sleepy) linear
```
sowasred2012: missed this yesterday - `animation` should appear before `opacity` in these three classes

waynecrosby:
The spacing between the product boxes has gone large again, the first set are ok but the ones below aren't. Also, is there a way to take over the scroll a bit with some inertia so that as you scroll down a section and near the end it almost snaps to the next section without being too aggressive and annoying?
Can you also add some padding to the bottom of the Partner section to extend the white down a bit.

solazio:
> The spacing between the product boxes has gone large again, the first set are ok but the ones below aren't. Also, is there a way to take over the scroll a bit with some inertia so that as you scroll down a section and near the end it almost snaps to the next section without being too aggressive and annoying?
> Can you also add some padding to the bottom of the Partner section to extend the white down a bit.

@waynecrosby I fixed the spacing and added some additional padding to the bottom of partners section. For the scrolling effect I created a new issue https://github.com/canonical-web-and-design/web-squad/issues/1532

PR #37, opened 2019-09-05
1586 add jobs feed for all roles

solazio:
## Done
- Add country list to the `send your CV` tab on `/careers/tech-ops` page
- Add job feed to all `/careers` pages

## QA
- Go to [/careers/tech-ops](http://0.0.0.0:8002/careers/tech-ops) and make sure the country select has all the countries
- Go to [/careers/tech-ops](http://0.0.0.0:8002/careers/tech-ops) and click on available roles
- Make sure you get a number of roles in the feed
- Click on one job and check if you can see job details

P.S. I need to refactor the templates used, but that will happen in a separate PR

## Issue

https://github.com/canonical-web-and-design/web-squad/issues/1586


templates/careers/jobs/job-detail.html
```diff
@@ -30,16 +30,21 @@ <h1>Careers</h1>
...
+      <h2 class="u-no-margin--bottom u-sv1">{{ job.title }}</h2>
+      <p class="p-muted-heading">{{ job.location }}</p>
     </div>
   </div>
   <div class="row">
+    <div class="col-12">
+      {{ job.content | safe }}
+    </div>
+  </div>
+  <!-- <div class="row">
```
nottrobin: Is this still here because you're going to uncomment and implement it in a future PR?

In general, we shouldn't leave commented-out code hanging around, 'cos it'll always be in version control history anyway if we really need it.

templates/careers/careers_base.html
```diff
@@ -88,7 +88,7 @@
       </select>
     </div>
     <div class="col-9 col-start-large-5">
-      {% if (request.path != "/careers/lifestyle" and request.path != "/careers/ethics" and request.path != "/careers/travel" and request.path != "/careers/progression" and request.path != "/careers/diversity" and request.path != "/careers/job-detail") %}
+      {% if (request.path != "/careers/lifestyle" and request.path != "/careers/ethics" and request.path != "/careers/travel" and request.path != "/careers/progression" and request.path != "/careers/diversity" and request.path != "/careers/jobs/<path>") %}
```
nottrobin: Didn't you change the `/careers/jobs` path?

nottrobin: Anyway, this conditional be neater with something like:

``` python3
{% if request.path not in ["/careers/lifestyle", "/careers/ethics", ...] %}
```

solazio: Yes, I did, but forgot to change here.

templates/careers/jobs/job-detail.html
solazio: Yes, the idea was to use some different type of design for job details, however this needs to be discussed further. I will remove it from now as we have it available in version control history, as you said.

templates/careers/all.html
```diff
@@ -0,0 +1,44 @@
...
+    </div>
+    <div class="row">
+      <div class="col-12">
+        <hr />
+      </div>
+    </div>
+    <div class="row">
+      <div class="col-12">
+        <p class="p-muted-heading">
+          <span style="text-transform: capitalize">Austin&nbsp;&#124;&nbsp;Beijing&nbsp;&#124;&nbsp;Boston&nbsp;&#124;&nbsp;London&nbsp;&#124;&nbsp;Taipei</span>
```
nottrobin: Are these characters URL-encoded for a reason? 'Cos you're trying to explicitly prevent wrapping?

templates/careers/careers_base.html
nottrobin: I think you missed my second comment:

> Anyway, this conditional be neater with something like:
> 
> ``` python3
> {% if request.path not in ["/careers/lifestyle", "/careers/ethics", ...] %}
> ```

webapp/app.py
```diff
@@ -38,4 +66,10 @@ def convert_to_kebab(kebab_input):
     words = re.findall(
         r"[A-Z]?[a-z]+|[A-Z]{2,}(?=[A-Z][a-z]|\d|\W|$)|\d+", kebab_input
     )
+
     return "-".join(map(str.lower, words))
+
+
+def remove_special_chars(text):
```
nottrobin: It looks like you added this function twice - once here and once in `get_job_feed.py`. If you need it here you should import it from there.

webapp/get_job_feed.py
```diff
@@ -0,0 +1,60 @@
...
+base_url = "https://api.greenhouse.io/v1/boards/Canonical/jobs"
+
+
+def get_job_feed(endpoint):
+    return api_session.get(f"{base_url}{endpoint}")
+
+
+def get_vacancies(department):
+    feed = get_job_feed("?content=true").json()
+    path_department = remove_special_chars(department)
```
nottrobin: Why is `remove_special_chars` necessary? What cases are we controlling for here?

webapp/app.py
```diff
@@ -18,13 +23,36 @@
...
+    return flask.render_template(
+        f"careers/{department}.html", vacancies=vacancies
+    )
+
+
+@app.route("/careers/<department>/<job_id>")
+def job_details(department, job_id):
+    job = get_vacancy(job_id)
+
+    if remove_special_chars(job["department"]) != remove_special_chars(
```
nottrobin: You might also want to `.lower()` both strings for a more inclusive comparison.

webapp/get_job_feed.py
```diff
@@ -0,0 +1,60 @@
...
+
+api_session = CachedSession(
+    fallback_cache_duration=300, file_cache_directory=".webcache"
+)
+
+
+base_url = "https://api.greenhouse.io/v1/boards/Canonical/jobs"
+
+
+def get_job_feed(endpoint):
```
nottrobin: I don't think one-line functions are particularly helpful (or even 3-line functions in most cases). This is taking DRY too far - you're obfuscating useful logic for the sake of saving at most one line of duplication.

templates/careers/all.html
solazio: Yes, I don't want them to wrap.

webapp/get_job_feed.py
solazio: I changed it to remove hyphens only.

PR #38, opened 2019-09-10
Add apply for a general job functionality

solazio:
## Done
- Add apply for a general job functionality from the `send your CV` tab on any `careers` page

## QA
- Go to [/careers/all](http://0.0.0.0:8002/careers/all) and click on the `Send your CV` tab
- Fill in all the fields, except the ones after CV attachment and click `Send now` button at the bottom of the page
- Check if the job application was successful on greenhouse

P.S. Ask @solazio for API_KEY in order to get this to work

## Issue

https://github.com/canonical-web-and-design/web-squad/issues/1605


webapp/app.py
```diff
@@ -30,10 +35,34 @@ def index():
 
 
 # Career departments
-@app.route("/careers/<department>")
+@app.route("/careers/<department>", methods=["GET", "POST"])
 def department_group(department):
     vacancies = get_vacancies(department)
 
+    if flask.request.method == "POST":
```
nottrobin: It seems rather over-complicated to submit people back to `request.path`, and then have this fork of this view function which processes data completely differently and then loads a completely different template.

If I were you I would instead create a single `@app.route("/careers/thank-you", methods=["POST"])`, and put all the submission functionality in there.

templates/partial/_careers-send-cv.html
```diff
@@ -3,19 +3,19 @@
     <div class="col-12">
       <h2>Can't find the ideal role for you?</h2>
       <p>Don't worry if you can't find the a role that fits you, send us your CV and we'll keep it on file to match any future roles to your skills to see if we can get a great match.</p>
-      <form>
+      <form action="{{request.path}}" method="POST" enctype="multipart/form-data">
```
nottrobin: I would change `{{ request.path }}` to `/careers/thank-you` - see https://github.com/canonical-web-and-design/canonical.com/pull/38/files#r323151734

webapp/app.py
nottrobin: Oh wait the template is the same, so you get returned to the list of careers for that department. Interesting... Maybe this is the best structure?

webapp/greenhouse_api.py
```diff
@@ -51,6 +55,34 @@ def get_vacancy(job_id):
     return job
 
 
+def submit_to_greenhouse(form_data, form_cv, job_id="1383152"):
+    # Encode the API_KEY to base64
+    auth = (
+        "Basic " + str(base64.b64encode(API_KEY.encode("utf-8")), "utf-8")[:-2]
```
nottrobin: Currently, the error I see if I don't provide the API key is:

```
AttributeError: 'NoneType' object has no attribute 'encode'
```

If you simply make this `os.environ["GREENHOUSE_API_KEY"]` then we'd get a nicer error like:

```
KeyError: 'GREENHOUSE_API_KEY'
```

webapp/app.py
solazio: @nottrobin in this case I will leave it as it is, unless you foresee any issues in the future.

webapp/greenhouse_api.py
```diff
@@ -9,8 +13,8 @@
     fallback_cache_duration=300, file_cache_directory=".webcache"
 )
 
-
-base_url = "https://api.greenhouse.io/v1/boards/Canonical/jobs"
+base_url = "https://boards-api.greenhouse.io/v1/boards/Canonical/jobs"
+API_KEY = os.environ["GREENHOUSE_API_KEY"]
```
nottrobin: Doing this here means the whole app will fail to run without this key - which is why the "test-site" CI test is failing, and it will also prevent demos from starting.

I suggested you do this at line 61 precisely because then you'd only see this error when you tried to load that specific page.

Personally I'm also not a big fan in general of keeping things in global constants when those things are used in only one or two places, so I'd prefer that you read the environment variable directly in the view function from a style perspective as well.

PR #40, opened 2019-09-24
Add careers modal to add/remove jobs you want to apply for

solazio:
## Done
- Add careers modal to add/remove jobs you want to apply for

## QA
- Go to [/careers/apply](http://0.0.0.0:8002/careers/apply)
- Click on`Add more roles` button and see the modal is displayed
- Click `X` button in the top left corner to close it
## Issue

https://github.com/canonical-web-and-design/web-squad/issues/1604

### DIFF
--- a/static/sass/_pattern_careers-modal.scss
+++ b/static/sass/_pattern_careers-modal.scss
@@ -1,0 +1,60 @@
+@mixin canonical-p-careers-modal {
+  #contact-modal {
+    // sass-lint:disable-line no-ids
+    &.p-modal {
+      align-items: start;
+      position: fixed;
+      z-index: 100;
+
+      input[type="checkbox"] + label {
+        margin-bottom: 0;
+        margin-top: 0;
+      }
+
+      .p-modal__dialog {
+        margin-bottom: 0;
+
+        @media (min-width: $breakpoint-medium + 100px) {
+          max-width: $breakpoint-medium;
+          min-width: $breakpoint-medium;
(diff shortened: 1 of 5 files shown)


templates/careers/apply.html
```diff
@@ -45,7 +45,7 @@ <h3>Roles applying for:</h3>
             <label for="CheckMe3">Software Engineer - MAAS</label>
           </div>
           <div class="col-3 u-align--right">
-            <button class="p-button has-icon">
+            <button class="p-button has-icon" onclick="document.querySelector('.p-modal').classList.remove('u-hide');">
```
anthonydillon: This would be better to call a function so you can null check the modal in case someone removes/edits it in the future it will error.

static/sass/_pattern_careers-modal.scss
```diff
@@ -0,0 +1,35 @@
+@mixin canonical-p-careers-modal {
+  #contact-modal {
+    // sass-lint:disable-line no-ids
+    &.p-modal {
+      align-items: start;
+      position: fixed;
+      z-index: 100;
```
anthonydillon: I think the modal is better appearing in the centre of the screen. I don't think you need these styles.

waynecrosby:
* Can we add a + to the 'Add more roles' button
* The dark grey behind the modal window doesn't cover the whole screen in Chrome, there is a right and bottom white margin. When you scroll it also stays at the top.

PR #41, opened 2019-09-26
Add filters & sorting to career pages

solazio:
## Done
- Add filters to all careers pages

## QA
- Go to [/careers/all](http://0.0.0.0:8002/careers/all) and click on the `Availabl roles` tab
- Change `Filter by` from the top of the page and check if the jobs are filtered
- Change `Sort by` and make sure the list is sorted accordingly 

## Issue

Fixes https://github.com/canonical-web-and-design/web-squad/issues/1511


templates/partial/_careers-available-roles.html
```diff
@@ -6,15 +6,14 @@
...
-            <option value="">Management</option>
-            <option value="">Senior roles</option>
-            <option value="">Full-time</option>
-            <option value="">Part-time</option>
+          <label for="filter" aria-label="Filter by job type" class="p-form__label">Filter by</label>
+          <select class="js-submit-on-change p-form__control" id="filter" name="content" aria-label="Filter by job type">
+            <option {% if not category %}selected="selected"{% endif %} value="{{request.path}}#available-roles">All</option>
+           <option {% if category and category == 'home-based' %}selected="selected"{% endif %} value="{{request.path}}?category=home-based#available-roles">Home based</option>
+           <option {% if category and category == 'office-based' %}selected="selected"{% endif %} value="{{request.path}}?category=office-based#available-roles">Office based</option>
+           <option {% if category and category == 'management' %}selected="selected"{% endif %} value="{{request.path}}?category=management#available-roles">Management</option>
```
nottrobin: Why do you need to put the full URL in the `value`? Wouldn't just the query string suffice? I believe `?something=else` should be a valid `href` that a browser would follow.

templates/partial/_careers-available-roles.html
```diff
@@ -36,7 +35,7 @@
     <div class="col-12">
       <form action="">
         <ul class="p-list--divided">
-        {% for job in vacancies %}
+        {% for job in vacancies|sort(attribute='date', reverse = True) %}
```
nottrobin: I think you should do the default sorting in `get_vacancies`:

``` python3
    return sorted(vacancies, key=lambda item: item['date'], reverse=True)
```

webapp/greenhouse_api.py
```diff
@@ -38,8 +43,25 @@ def get_vacancies(department):
...
+        elif category == "office-based":
+            return filter_jobs(vacancies, "office", "Office Based")
+        elif category == "full-time":
+            return filter_jobs(vacancies, "employment", "Full-time")
+        elif category == "part-time":
+            return filter_jobs(vacancies, "employment", "Part-time")
+        elif category == "management":
+            return filter_jobs(vacancies, "management", True)
+
     return vacancies
```
nottrobin:
```suggestion
    return sorted(vacancies, key=lambda item: item['date'], reverse=True)
```

nottrobin:
`#available-roles` doesn't seem to switch back to the tab on reload

nottrobin:
Actually, I'd be tempted to do the filtering client-side as well - given that presumably the initial load state will always be "all roles", so the client already has the data to do any filtering.

nottrobin:
Should there be an empty state message?

image

PR #46, opened 2019-10-07
Add filter and search to find-a-partner page

solazio:
## Done
- Add search functionality to `find-a-partner` page
- Add filter functionality to `find-a-partner` page

## QA
- Go to [/partners/find-a-partner](http://0.0.0.0:8002/partners/find-a-partner) and search for partners and filter partners 
- Make sure it behaves similar to https://partners.ubuntu.com/find-a-partner

## Issue
Fixes https://github.com/canonical-web-and-design/web-squad/issues/1645


nottrobin:
In testing on Firefox, the Country filter didn't seem to do anything, neither did the Technology filter. The Services Offered seems to work in some cases but not others - if I tick "Hosting provider" I seem to get partners that don't appear to be hosting providers.

solazio:
> In testing on Firefox, the Country filter didn't seem to do anything, neither did the Technology filter. The Services Offered seems to work in some cases but not others - if I tick "Hosting provider" I seem to get partners that don't appear to be hosting providers.

@nottrobin I should have mentioned that in the PR, the country filter doesn't work as there is no info in the api related to the location. I will have another look ar Technology and Hosting providers. Thanks.

solazio:
@nottrobin tested it again on both ff and chromium and all filters seem work and  display the correct information.

static/js/find-a-partner.js
```diff
@@ -0,0 +1,170 @@
+(function() {
+
+  var filters = [];
+
+  function init() {
+    revealSearch();
+
+    var inputNode = document.querySelector(".p-find-a-partner__search-input");
+    inputNode.addEventListener("keyup", searchHandler);
```
anthonydillon: You probably want to null check this as it will error if someone changes the class. Can you also use `js-` prefixed classes so its clear they apply functionality and no styling.

static/js/find-a-partner.js
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  function matchesExist() {
+    var numberOfPartners = document.querySelectorAll(
+      ".p-find-a-partner__partner"
+    ).length;
+    var numberOfHidden = document.querySelectorAll(
+      ".p-find-a-partner__partner.u-hide"
+    ).length;
+    if (numberOfHidden === numberOfPartners) {
```
anthonydillon: You could just check if any are hidden. In doing so, removing the `numberOfPartners` query and just checking if `numberOfHidden` is greater then 0.

static/js/find-a-partner.js
```diff
@@ -0,0 +1,170 @@
...
+      return false;
+    } else {
+      return true;
+    }
+  }
+
+  function updateNoResultsMessage() {
+    var noResults = document.querySelector(".p-find-a-partner__no-results");
```
anthonydillon: You should null check for this element and use `js-` classes to target it.

static/js/find-a-partner.js
```diff
@@ -0,0 +1,170 @@
...
+          checkboxObject.setAttribute("checked", "checked");
+          updateFilter(key + "-" + queryParams[key][i], true);
+          document.querySelector("#" + key).classList.add("open");
+        }
+      }
+    }
+  }
+
+  function populateTextbox() {
+    var searchbox = document.querySelector(".p-find-a-partner__search-input");
```
anthonydillon: Null check again

static/js/find-a-partner.js
```diff
@@ -0,0 +1,170 @@
...
+    var searchText = getParameterByName("search");
+    if (searchbox && searchText) {
+      searchbox.focus();
+      searchbox.value = searchText;
+      search(searchText);
+    }
+  }
+
+  function revealSearch() {
+    var searchForm = document.querySelector(".p-strip--light.is-shallow.u-hide");
```
anthonydillon: Target a js class please

anthonydillon: If someone changes the design the functionality here will break

templates/partners/find-a-partner.html
```diff
@@ -399,53 +399,43 @@ <h1>Find a Partner</h1>
...
+            </p>
+            <p class="p-card__content">{{ partner['short_description']|markup|safe }}</p>
+          </div>
         </div>
+        {% endfor %}
       </div>
+      {% endif %}
     </div>
   </section>
+  <script async defer src="/static/js/find-a-partner.js"></script>
```
anthonydillon: This has both defer and async. You just want defer.

anthonydillon:
QA:
- Once searched the clear X clears the search box but doesnt update the partners list
- Might be nice to display the number of partners and the search term in small text under the search box
- The ARM logo is too small, is this all coming from the API?
- I'm not 100% sure but seems if I search for say "dd" and then filter by cloud the search is ignored and the filter is applied. I would expect both to apply

solazio:
@anthonydillon I added the empty state message and now will merge.

PR #48, opened 2019-10-08
Implement initial basic navigation

solazio:
## Done
- Add basic navigation

## QA
- Go to [/partners/find-a-partner](http://0.0.0.0:8002/partners/find-a-partner) and check if the navigation works

## Issue
Fixes https://github.com/canonical-web-and-design/web-squad/issues/1769

### DIFF
--- a/static/sass/_pattern_navigation.scss
+++ b/static/sass/_pattern_navigation.scss
@@ -1,0 +1,18 @@
+@mixin canonical-p-navigation {
+  .p-navigation {
+    background-color: #2c001e;
+
+    .p-navigation__links > .p-navigation__link > a {
+      color: $color-x-light;
+
+      &:hover {
+        color: $color-mid-dark;
+      }
+    }
+
+    .p-navigation__link.is-selected > a {
+      background-color: $color-x-light;
+      color: $color-x-dark;
+    }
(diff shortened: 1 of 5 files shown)


static/sass/_pattern_navigation.scss
```diff
@@ -0,0 +1,18 @@
+@mixin canonical-p-navigation {
+  .p-navigation {
+    background-color: #2c001e;
+
+    .p-navigation__links > .p-navigation__link > a {
+      color: $color-x-light;
+
+      &:hover {
+        color: $color-mid-dark;
```
sowasred2012: Against the aubergine background, this hover state color fails the color contrast ratio accessibility test: https://webaim.org/resources/contrastchecker/?fcolor=666666&bcolor=2C001E

It's a hover state, so there's an argument to be made that any user with accessibility needs has likely hovered over the element because they intend to click on it, but there's a chance that isn't the case and [the guidance](https://webaim.org/articles/contrast/#changes) recommends maintaining an accessible color contrast in all states.

sowasred2012:
Sorry, just noticed this too on small screens:

Screenshot 2019-10-08 at 17 17 36

PR #53, opened 2019-10-14
WIP: 1664 implement multiple job application

solazio:
## Done
- Change job details route to `/careers/{job_id}`
- Add empty state message to all `careers` pages
- Add functionality to apply for multiple jobs

## QA
- Go to [/careers/start](http://0.0.0.0:8002/careers/start) 

## Issue
Fixes https://github.com/canonical-web-and-design/web-squad/issues/1664
Fixes https://github.com/canonical-web-and-design/canonical.com/issues/42


webapp/app.py
```diff
@@ -96,17 +96,11 @@ def department_group(department):
     )
 
 
-@app.route("/careers/<department>/<job_id>")
-def job_details(department, job_id):
+@app.route("/careers/job/<job_id>")
```
tbille: is that a url that we want? `/careers/job` sounds a bit redoundant no?

tbille: the job id might be something unique. (`/careers/job/1` for example) so a url that we will never use. Why not just putting `/careers/1` feels shorter and cleaner URL.

tbille: even what would be great would be to have the slug instead of the number but I don't know if greenhouse provides this.

webapp/app.py
```diff
@@ -96,17 +96,11 @@ def department_group(department):
...
-        and "all" != department.lower()
-    ):
-        flask.abort(404)
-
-    return flask.render_template("/careers/jobs/job-detail.html", job=job)
+
+    return flask.render_template("/careers/job/job-detail.html", job=job)
 
 
 @app.route("/careers/<department>/<job_id>", methods=["POST"])
```
tbille: also here to make it more coherent I would use the same URL system.

`GET,POST -> /careers/job/<id>` or whatever url is used

webapp/app.py
solazio: @tbille greenhouse only takes id to identify the job. To be honest I'm not against using `/careers/1` instead of `/careers/job/1`, but I thought `/careers/job/1` it makes more sense as it makes it clear you are looking at a job.

tbille: >  greenhouse only takes id to identify the job. 

ok

> To be honest I'm not against using /careers/1 instead of /careers/job/1, but I thought /careers/job/1 it makes more sense as it makes it clear you are looking at a job.

Well when you are in the `/careers/*` it already implies that you are looking for a job.

solazio: @tbille will change it to `/careers/1` as per your suggestion. thanks

webapp/partners_api.py
```diff
@@ -13,92 +14,53 @@ def get(endpoint):
...
 
 
-def first_ten(req):
-    new_req = req[:10]
+def random_ten(req):
+    if len(req) >= 10:
+        new_req = sample(req, 10)
+    else:
+        new_req = sample(req, len(req))
+    print(len(new_req))
```
tbille: remove print

webapp/partners_api.py
```diff
@@ -13,92 +14,53 @@ def get(endpoint):
...
     return new_req
 
 
 def get_partner_groups():
     partner_groups = {
```
tbille: we have to make one api call?

Here if one fails the entire page will fail.

webapp/partners_api.py
```diff
@@ -1,5 +1,6 @@
 from canonicalwebteam.http import CachedSession
 
+from random import sample
```
tbille:
```
from random import sample

from canonicalwebteam.http import CachedSession
```

webapp/greenhouse_api.py
```diff
@@ -90,32 +90,52 @@ def get_vacancy(job_id):
...
+            # "educations": form_data[educations],
         }
     )
+    # print(payload)
 
     response = requests.post(
         f"{base_url}/{job_id}", data=payload, headers=headers
     )
-
+    print(response)
```
tbille: remove print

webapp/app.py
```diff
@@ -63,10 +63,57 @@ def results():
...
+    if len(messages) > 0:
+        return flask.render_template(
+            "careers/thank-you.html", messages=messages
+        )
+    else:
+        return flask.redirect("/careers")
+
+
+@app.route(
+    "/careers/<regex('[a-z-]*[a-z][a-z-]*'):department>",
```
tbille: I don't think this is what you are searching for.

What is the format of the departemnts?

webapp/app.py
```diff
@@ -63,10 +63,57 @@ def results():
...
 
-@app.route("/careers/<department>", methods=["GET", "POST"])
+@app.route("/careers/thank-you", methods=["POST"])
+def careers_thank_you():
+    messages = []
+    job_id_list = flask.request.form.get("applicationJobIdList").split(",")
+    job_title_list = flask.request.form.get("applicationJobTitleList").split(
+        ","
+    )
+    print(flask.request.form)
```
tbille: print

webapp/app.py
```diff
@@ -63,10 +63,57 @@ def results():
...
+    i = 0
+    while i < len(job_id_list):
```
tbille: no for loops like this.

`for job_id in job_id_list:`

webapp/app.py
```diff
@@ -63,10 +63,57 @@ def results():
     return flask.render_template("careers/results.html", **context)
 
 
-@app.route("/careers/<department>", methods=["GET", "POST"])
+@app.route("/careers/thank-you", methods=["POST"])
+def careers_thank_you():
+    messages = []
+    job_id_list = flask.request.form.get("applicationJobIdList").split(",")
```
tbille: job_ids

webapp/app.py
```diff
@@ -63,10 +63,57 @@ def results():
...
+    job_title_list = flask.request.form.get("applicationJobTitleList").split(
```
tbille: job_titles

webapp/app.py
```diff
@@ -63,10 +63,57 @@ def results():
     return flask.render_template("careers/results.html", **context)
 
 
-@app.route("/careers/<department>", methods=["GET", "POST"])
+@app.route("/careers/thank-you", methods=["POST"])
```
tbille: no post to thank you page. It doesn't makes sense. Keepp the POST how it was and then redirect to `/careers/thank-you`

webapp/app.py
```diff
@@ -63,10 +63,57 @@ def results():
...
+                }
+            )
+        else:
+            messages.append(
+                {
+                    "type": "error",
+                    "title": f"{job_title_list[i]}",
+                    "text": (
+                        f"Error {response.status_code}. {response.reason}."
+                        "<br /><a href='/careers/{job_id_list[i]}' "
```
tbille: no html in backend.
`"style='padding-left: 2rem;'>"` this will be impossible to maintain

webapp/app.py
```diff
@@ -63,10 +63,57 @@ def results():
...
+                        "style='padding-left: 2rem;'>"
+                        "Please try to apply to this job again!</a>"
+                    ),
+                    "response": response,
+                }
+            )
+        i = i + 1
+        # time.sleep(60)
```
tbille: comment

webapp/partners_api.py
solazio: @tbille I agree that one api call would be better, however, in this situation if I make one api call and get all the partners it takes a long time (like 4-5 seconds). I can definitely do that , but let's catch up next week and discuss further about this.

steverydz:
It doesn't look like this has been worked on in a while and is quite out of date. Can it be closed?

solazio:
@steverydz yes. Closing...

PR #62, opened 2019-10-17
Add empty state for careers pages

solazio:
## Done
- Add empty state message to all `careers` pages

## QA
- Go to [/careers/engineering](http://0.0.0.0:8002/careers/engineering) 

## Issue
Fixes https://github.com/canonical-web-and-design/canonical.com/issues/42

### DIFF
--- a/templates/partial/_careers-available-roles.html
+++ b/templates/partial/_careers-available-roles.html
@@ -1,4 +1,5 @@
 <div class="p-strip is-shallow u-extra-space tab-content u-hide" id="available-roles">
+  {% if vacancies %}
   <div class="row u-vertically-center">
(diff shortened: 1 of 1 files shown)


templates/partial/_careers-available-roles.html
```diff
@@ -35,6 +35,7 @@
       <hr />
     </div>
   </div>
+  {% if vacancies %}
```
sowasred2012: Could this go on line 2 instead? If there are no vacancies, there's nothing to filter or sort.

solazio: @sowasred2012 you are right.

PR #67, opened 2019-10-21
Add careers filters to the url

solazio:
## Done
- Change the url when filter is changed on `/careers/{department}#available-roles`
- Add empty state message if no jobs match filter selection

## QA
- Go to [/careers/sales#available-roles](http://0.0.0.0:8002/careers/sales#available-roles) and change the filter on the page. See that the url is changed when you do that.
- Make sure there is an empty state message if no matches for your filter.

## Issue
Fixes https://github.com/canonical-web-and-design/canonical.com/issues/43


solazio:
@sowasred2012 the link between url ids and tabs is the next thing I need to implement. I used ids in the url with that in mind.

PR #122, opened 2019-12-18
Parse the feed departments to work with the site

anthonydillon:
## Done
Added a parsing function to the returned feed department data to align with the site content.

## QA
- View the site locally in your web browser at: http://0.0.0.0:8002/careers
- Click through and check to engineer and we and design have roles available
- The others do not yet
- Compare to the `department - external` here: https://boards-api.greenhouse.io/v1/boards/Canonical/jobs

### DIFF
--- a/webapp/greenhouse_api.py
+++ b/webapp/greenhouse_api.py
@@ -22,6 +22,7 @@ def get_vacancies(department):
     vacancies = []
     for job in feed["jobs"]:
         feed_department = remove_hyphens(job["metadata"][2]["value"])
+        feed_department = parse_feed_department(feed_department)
         if path_department.lower() == "all":
             vacancies.append(
                 {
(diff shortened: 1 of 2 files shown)


webapp/greenhouse_api.py
```diff
@@ -22,6 +22,7 @@ def get_vacancies(department):
     vacancies = []
     for job in feed["jobs"]:
         feed_department = remove_hyphens(job["metadata"][2]["value"])
```
hatched: I'm not a fan of variable re-assignment. I'd prefer something like...
```suggestion
        feed_department = parse_feed_department(
            remove_hyphens(job["metadata"][2]["value"]))
```

webapp/greenhouse_api.py
```diff
@@ -137,6 +138,21 @@ def submit_to_greenhouse(form_data, form_files, job_id="1658196"):
     return response
 
 
+def parse_feed_department(feed_department):
+    if (
```
hatched: I kind of find this hard to read, maybe a `switch` or a `map` would be more appropriate.

hatched:
QA Issue: 

If I navigate to `https://canonical-com-canonical-web-and-design-pr-122.run.demo.haus/careers/engineering#available-roles` then hit the back button to return to `/careers/engineering` the content of the tabs are empty.

webapp/greenhouse_api.py
anthonydillon: Done

webapp/greenhouse_api.py
anthonydillon: Done

anthonydillon:
@hatched the QA issue is due to the use of JS when CSS should do the trick. I'll refactor the tabs to be CSS only in a follow up branch

PR #138, opened 2020-01-06
Fix filters on results

### LINKED ISSUE #125: "Filter by" doesn't appear to work on career game results
When completing the careers game and viewing the results, it looks like the "Filter by" dropdown doesn't work. Doesn't seem to happen when viewing jobs by the categories listed in the navigation.

careers

sowasred2012:
## Done

- fixed dropdown filters on /careers/results
- fixed issue where refreshing the page after using the dropdown filter would cause the 'coreskills' values to be lost

## QA

- View the site locally in your web browser at: http://0.0.0.0:8002/careers/start
- select some options, submit them - you should land on /careers/results
- using the dropdown, filter the results
- see that the results are indeed filtered
- refresh the page, see that the page loads successfully, and the relevant results are shown

## Issue / Card

Fixes #125

### DIFF
--- a/templates/careers/_filters.html
+++ b/templates/careers/_filters.html
@@ -14,10 +14,10 @@
   <div class="p-form__group">
     <label for="sort" aria-label="Sort resources" class="p-form__label">Sort by</label>
     <select class="js-sort p-form__control" id="sort" name="sort" aria-label="Sort resources">
-      <option>Date</option>
-      <option>Location</option>
+      <option value="date">Date</option value="">
+      <option value="location">Location</option>
       {% if request.path == "/careers/all"%}
-      <option>Job sector</option>
+      <option value="sector">Job sector</option>
       {% endif %}
     </select>
   </div>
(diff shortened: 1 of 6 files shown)


templates/careers/_filters.html
```diff
@@ -0,0 +1,26 @@
...
+      <option value="office-based">Office based</option>
+      <option value="management">Management</option>
+      <option value="full-time">Full-time</option>
+      <option value="part-time">Part-time</option>
+    </select>
+  </div>
+  <div class="p-form__group">
+    <label for="sort" aria-label="Sort resources" class="p-form__label">Sort by</label>
+    <select class="js-sort p-form__control" id="sort" name="sort" aria-label="Sort resources">
+      <option>Date</option>
```
steverydz: I think these should have values.

sowasred2012: Done - tweaked the JS that was using this too.

PR #195, opened 2020-03-31
Replace canonicalwebteam.http with Talisker session

nottrobin:
This removes timeouts on requests, which hopefully fixes https://github.com/canonical-web-and-design/canonical.com/issues/194.

A bit more detail here: https://github.com/canonical-web-and-design/canonical.com/issues/194#issuecomment-606610169

QA
--

Test the site, check tests pass.


webapp/app.py
```diff
@@ -76,11 +75,13 @@ def results():
...
 def department_group():
     department = flask.request.path.split("/")[2]
-    vacancies = get_vacancies(department)
+    vacancies = greenhouse_api.get_vacancies(department)
 
     if flask.request.method == "POST":
-        response = submit_to_greenhouse(
-            flask.request.form, flask.request.files
+        response = greenhouse_api.submit_application(
+            os.environ["GREENHOUSE_API_KEY"],
```
tbille, line 82: Can this go into a varaible that gets instantiate on start of the app.

nottrobin: No because we don't want it to fail until it's needed. As in, you can run the site fine without this API key until you try to submit an application, then only it fails.

nottrobin: Oh alright I'll make it prettier

PR #273, opened 2020-08-19
Add dynamic navigation to careers page

### LINKED ISSUE #263: Dynamic careers navigation
We should load the external department's list from the Greenhouse API and render the side navigation from that. If the site has overview content for that department we should render the overview tab. If a new department is created or one renamed it should only show the available roles until the content has been created for the department.

There is no "easy" way to add a blob of content to a department in Greenhouse as it is simply a list.

MeltySnow:
## Done

- Dynamically generate the sidebar navigation on the canonical.com/careers pages from the external departments list on the Greenhouse Harvest API.
- Added generic template for careers pages without overview content.

## QA

- You will need to add a Harvest API key to your local `.env.local` file (create one if it does not exist) under the name HARVEST_API_KEY. If you need one, let me know and I'll send one to you.
- Run the site using the command `./run`
- View the site locally in your web browser at: http://0.0.0.0:8002/careers/all
- Ensure that the sidebar navigation contains all the fields that are currently in the live version
- Ensure that the vacancy counter in the sidebar is working properly and is suitably formatted
- Ensure that the pages with existing templates render properly (for example, all those which are currently on the live site), including those which are not department pages such as http://0.0.0.0:8002/careers/lifestyle
- Ensure that all relevant vacancies still show up properly
- Ensure that the default template is suitably formatted (at this time, I believe http://0.0.0.0:8002/careers/product-management is the only page that uses this template)
- Ensure that the error page for nonexistent departments (such as http://0.0.0.0:8002/careers/none) is suitable
- Ensure that the code is sensible, both semantically and in terms of formatting

## Issue / Card

Fixes #263


webapp/app.py
```diff
@@ -69,21 +67,93 @@ def results():
...
+    """
+    for item in all_departments:
+        new_department = Department(item)
+        is_new = True
+        for department in departments:
+            if department.slug == new_department.slug:
+                is_new = False
+                break
+        if is_new:
+            departments.append(new_department)
```
anthonydillon, line 114: I think this is good but could be improved by using a Set which de-dups automatically

webapp/app.py
```diff
@@ -69,21 +67,93 @@ def results():
...
+
+        if feed_department.lower() in field:
+            return field[feed_department.lower()]
+
+        return feed_department
+
+    def __init__(self, name):
+        self.name = name
+        self.slug = Department.__parse_feed_department(name)
+        self.slug = self.slug.replace(" ", "-").lower()
```
anthonydillon, line 94: This manipulation would be better suited in the parse function

webapp/app.py
```diff
@@ -69,21 +67,93 @@ def results():
...
-@app.route("/careers/tech-ops", methods=["GET", "POST"])
-def department_group():
-    department = flask.request.path.split("/")[2]
-    vacancies = greenhouse_api.get_vacancies(department)
+# Class that collects department-specific content
+class Department(object):
+    __careers_directory = "./templates/careers"
+
+    def __parse_feed_department(feed_department):
+        field = {
```
anthonydillon, line 75: I think `field_mapping` would be a better name here.

webapp/app.py
```diff
@@ -69,21 +67,93 @@ def results():
...
-@app.route("/careers/marketing", methods=["GET", "POST"])
-@app.route("/careers/project-management", methods=["GET", "POST"])
-@app.route("/careers/sales", methods=["GET", "POST"])
-@app.route("/careers/tech-ops", methods=["GET", "POST"])
-def department_group():
-    department = flask.request.path.split("/")[2]
-    vacancies = greenhouse_api.get_vacancies(department)
+# Class that collects department-specific content
+class Department(object):
+    __careers_directory = "./templates/careers"
```
anthonydillon, line 72: Doesn't look like this is used.

webapp/app.py
```diff
@@ -69,21 +67,93 @@ def results():
...
+    def __parse_feed_department(feed_department):
+        field = {
+            "cloud engineering": "engineering",
+            "device engineering": "engineering",
+            "web and design": "design",
+            "web & design": "design",
+            "operations": "commercial-ops",
+            "human resources": "hr",
+            "techops": "tech-ops",
+            "product": "product management"
```
anthonydillon, line 83: Are these still required? I was under the impression that the data source has been aligned?

webapp/app.py
```diff
@@ -69,21 +67,93 @@ def results():
...
+        vacancy_count[dept.slug] += 1
+
+        if dept.slug == department or department == "all":
+            vacancies.append(vacancy)
+
+    """
+    Generate list of templates in the /templates/careers folder,
+    and remove the .html suffix
+    """
+    for template in os.listdir("./templates/careers"):
```
anthonydillon, line 153: @nottrobin is this going to be to expensive a query on page view?

webapp/greenhouse.py
```diff
@@ -177,3 +179,16 @@ def submit_application(
         )
 
         return response
+
+    # Get list of external departments from the Harvest API
+    def get_departments(self):
+        content_url = "https://harvest.greenhouse.io/v1///custom_field/155450"
```
anthonydillon, line 185: Do you need the `///` in the URL?

anthonydillon: Can you call this what it is. Something like `department_api_url`

webapp/greenhouse.py
```diff
@@ -177,3 +179,16 @@ def submit_application(
...
+        key = harvest_api_key + ":"
```
anthonydillon, line 186: Can you add a comment here to explain why you are adding a `:` to the key? Something like, "The key is passed as the username with a blank password. Therefore, the key needs an appended colon to achieve this"

webapp/greenhouse.py
```diff
@@ -177,3 +179,16 @@ def submit_application(
...
+        auth = ("Basic " + str(base64.b64encode(key.encode("utf-8")), "utf-8"))
```
anthonydillon, line 187: This could be `base64_encoded_key`

anthonydillon (changes requested):
Some suggestions and questions in the code.

anthonydillon:
This is awesome @MeltySnow! Great work. I have found one QA issue and added a design label to get the count reviewed @waynecrosby could you take a look and see what you think. I'll add screenshots as there is no easy way to view them.

## QA
- There is a missing vacancy_count from the view when you complete the "game thing": http://0.0.0.0:8002/careers/start.

anthonydillon:
## Screenshots
image

image

anthonydillon:
I tried with [Vanilla labels](https://vanillaframework.io/docs/patterns/labels) but we only have colour ones for now. @lyubomir-popov 

image

anthonydillon:
Maybe only have it on the All roles at the bottom and add the label to the tab on Available roles:
image

webapp/app.py
MeltySnow: Some of these mappings are redundant but there are still a few anomalies; for instance the template corresponding to `web & design` is called `design.html`

lyubomir-popov:
@anthonydillon Why not just use:

- right aligned links if these need to be clickable
- right aligned muted text if not clickable?

webapp/app.py
MeltySnow: Amendment to my last comment: the Department container class is also used for comparing the `department` property of vacancies to the external departments list, so all these key:value pairs are necessary

webapp/app.py
MeltySnow: The departments list get indexed in some jinja templates, and to my knowledge sets don't support indexing?

waynecrosby:
@anthonydillon I thought the first was fine, but if you want to bring out the numbers stronger for that section then you could combine the 2...
opt1
opt2

MeltySnow:
I've pushed up a change that addressed a bunch of @anthonydillon's comments, including the QA issue. Since the results page has its own route (which is more specific and presumably has higher precedence than the generic one), it was being rendered without the variables needed for the navigation so I moved the navigation dependencies into their own function and had to change a bunch of variable names to avoid clashes.

anthonydillon:
@waynecrosby @lyubomir-popov I'm happy if you guys are. Would you mind giving it a design +1 in that case?

anthonydillon:
@nottrobin do you think we should build in exception handling for demos and and local devs that don't have the API key?

nottrobin:
@anthonydillon well, the page should certainly return a something more informative than a 500. I think maybe a "502 bad gateway" would be appropriate. And that error page could carry the message that the API key is missing.

But this shouldn't apply to demos - we should just give the demos the secret. Let me know if you want help with that.

nottrobin:
@anthonydillon, @MeltySnow: Ah the 500 on the demo [here](https://canonical-com-273.demos.haus/careers/start) isn't because of the secret at all, it's [a type error](https://sentry.is.canonical.com/canonical/canonical-com/issues/11610/?query=is%3Aunresolved).

MeltySnow:
@nottrobin I think the type error happens *because* there is no secret. Presumably in this line: https://github.com/canonical-web-and-design/canonical.com/blob/a0d8fa96d03495eda8aec15f10f8430dfdbfd4ea/webapp/greenhouse.py#L11 `os.environ.get("HARVEST_API_KEY")` returns `NoneType`, and then this line: https://github.com/canonical-web-and-design/canonical.com/blob/a0d8fa96d03495eda8aec15f10f8430dfdbfd4ea/webapp/greenhouse.py#L188 throws a type error because there is no `+` overload for `NoneType`.

webapp/app.py
anthonydillon: We could sort alphabetically each time right?

webapp/app.py
anthonydillon: Ok, just checking.

anthonydillon: We want to remove as much logic as possible from this middleware to reduce errors in the future. If that needs URLs to change and adding redirects that would be fine.

anthonydillon:
Yeah @nottrobin there is a new API key for the harvest API which we didn't have to use before.

webapp/app.py
anthonydillon: Sorry the function should remain `__parse_feed_department` its the object `field` that should be renamed `field_mapping`

anthonydillon (approved):
Great work @MeltySnow LGTM :+1:

anthonydillon:
We need to ensure that production has access to the API thought the firewall and that the API token has been rolled out across the environments before we merge and release this.

jkfran:
> We need to ensure the following before we merge and release this:
> 
>     * [x]  that production has access to the API thought the firewall
> 
>     * [x]  that the API token has been rolled out across the environments

Done

jkfran:
@nottrobin @anthonydillon I did some changes but tests are not passing because the Harvest API secret is not available on CircleCI. Should I proceed and mock the Harvest API responses for the tests?

nottrobin:
Based on no context, that certainly sounds like the right solution

jkfran:
Okay, can you review my last three commits and merge if it's fine?

anthonydillon (changes requested):
## QA issues:
- http://0.0.0.0:8002/careers/start is 404ing
- The job counts are incorrect, maybe due to them being hidden from public but still being counted:
image
- http://0.0.0.0:8002/careers/all is 404ing too

jkfran:
@anthonydillon I am debugging the code for the counts, I think the counts are fine the issue is getting the jobs for Web and design.

The following condition doesn't match because the values are "'web&design"=="design"`

https://github.com/canonical-web-and-design/canonical.com/pull/273/files#diff-9a62d0498494ef0060d333810f91c56fL49

I'll fix it

jkfran:
@anthonydillon 🙄

anthonydillon:
@jkfran thanks for fixing them.

There seems to be no roles for Product now: https://canonical-com-273.demos.haus/careers/product-management

anthonydillon (approved):
Code and QA :+1:

PR #308, opened 2020-11-02
Update cookie policy to 3.0.5

pmahnke:
## Done

- Add cookie policy module

## QA

- View the site locally in your web browser at: http://0.0.0.0:8002/partners
- Check you see the cookie policy module panel
- Open the manager and make a selection
- The panel should disappear and the `_cookies_accepted` cookie should be set
- See that the cookie's path is set to '/'

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "test-python": "python3 -m unittest discover tests"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


pmahnke:
caught by renovate

PR #370, opened 2021-03-25
Condense the layout of the jod details page

anthonydillon:
## Done
Condense the layout of the job details page

## QA
- Open the demo
- Go to /careers/2383894/vp-cloud-alliances-remote
- See that there is more content above the fold than [on live](https://canonical.com/careers/2383894/vp-cloud-alliances-remote)
- Check it looks ok on small screens

## Screenshots

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/templates/careers/job-detail.html
+++ b/templates/careers/job-detail.html
@@ -7,28 +7,22 @@
 {% block meta_keywords %}{{ job.skills | join(', ') }}{% endblock %}
 
 {% block hero %}
-<section class="p-strip--image is-dark is-deep u-no-padding--top u-no-padding--bottom u-extra-space">
-  <div class="p-strip--suru-background">
-    <div class="row">
-      <div class="col-6">
-        <h1>Careers</h1>
-      </div>
+<section class="p-strip--suru-topped u-no-padding--bottom">
+  <div class="row">
+    <div class="col-6">
+      <a href="{% if request.referrer %}{{ request.referrer }}{% else %}/careers/all{% endif %}" class="u-sv2">&lsaquo;&nbsp;Back to list</a>
     </div>
   </div>
+  <div class="row u-equal-height">
+    <div class="col-6">
+      <h1>{{ job.title }}</h1>
+      <p class="p-muted-heading">{{ job.location }}</p>
+    </div>
+    <!-- <div class="col-6 u-align--right u-vertically-center">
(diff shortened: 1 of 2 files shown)


nottrobin (approved):
Lgtm

templates/careers/job-detail.html
```diff
@@ -45,73 +39,60 @@ <h1>Careers</h1>
...
+
+  <div class="col-5 col-start-large-8 p-card--strip">
+    <h3>Apply for this role</h3>
+    <form action="{{ request.path }}" method="POST" enctype="multipart/form-data" class="js-roles-list--form" onsubmit="dataLayer.push({
+          'event': 'GAEvent',
+          'eventCategory': 'Form',
+          'eventAction': 'job application: {{ job.id }}',
+          'eventLabel': 'Submit Application',
+          'eventValue': undefined });" >
+      <fieldset style="border: none; margin-bottom: 0;">
```
carkod, line 58: `u-no-margin--bottom`? I am not sure if we have something for border maybe we should?

lyubomir-popov:
Is there a reason the fieldset is styled differently compared to the [vanilla default](https://vanillaframework.io/docs/examples/base/forms/fieldset)?

carkod:
Also when I enter a wrong format for email address it returns `Unprocessable entity` which I think it's not very friendly error message. Maybe we should put in the validation JS that we have on ubuntu.com? But I can file an issue for that if you prefer

waynecrosby:
@anthonydillon Can we make the title go the full 12cols width, it's still wrapping in the left hand column so for long titles it knocks the form down?

image

anthonydillon:
> Is there a reason the fieldset is styled differently compared to the vanilla default?

@lyubomir-popov totally agree but I didn't change that just moved its position.

templates/careers/job-detail.html
anthonydillon: Ideally we would code up a true pattern in Vanilla but that will be in the future.

anthonydillon:
@waynecrosby good point! Done.
image

waynecrosby:
@anthonydillon @lyubomir-popov Looks good. The form was originally designed to be like the below, I am assuming it's because its pulled from Greenhouse into our page that we get the strange box-in-box effect? Happy to revert to the simple Vanilla standard white box with outline.

image

PR #392, opened 2021-06-09
Update docker image from node-slim to node

tbille:
## Done

Update docker image from node-slim to node since some node dependencies require python that node slim doens't integrate

## QA

make sure demo builds


pmahnke (approved):
thanks

PR #402, opened 2021-07-15
Fix "All roles" count

### LINKED ISSUE #396: all roles counter shows 0
hi web team!
looks like our all roles counter is showing 0, which is incorrect
Screenshot from 2021-07-02 14-32-39

mrgnr:
## Done

Fixed the count next to the "All roles" link, which was previously always showing 0.

## QA

- Run the site using the command `dotrun`
- View the site locally in your web browser at: http://0.0.0.0:8002/careers/all
- Check that the count shown next to the "All roles" link in the careers navigation sidebar is not zero (it should be equal to the sum of the count for each department)

## Issue / Card

Fixes #396 

## Screenshots

Screenshot from 2021-07-15 17-33-30


anthonydillon (approved):
Nice work @mrgnr  :+1:

PR #404, opened 2021-07-20
Move skip link above main navigation

sowasred2012:
## Done

Moved the skip link above the main nav so it's the first focussable element on the page

## QA

- Press the tab key, if the focus isn't already on the page, it should reveal the skip link on the first press


bethcollins92 (approved):
ahh makes sense, LGTM!

PR #428, opened 2021-09-22
Replaced the field "department" with "departments"

### LINKED ISSUE #352: [c.c career enhancements] Add ability for roles to have multiple departments in greenhouse and use on c.c
[Brief](https://docs.google.com/document/d/1eXlLZKLmq_tLrOCgulbT4KYhsCq3FZ8AA7KX7L2jMeY/edit#)

We would have to create a new custom field and then migrate the existing field to the multiselect one and update canonical.com to use this instead of the external department field (you cannot migrate the field type).

goulinkh:
## Done

- Updated the structure of Vacancy on the backend side
- Updated the templates where there is a reference to vacancy.department

*FYI: Departments is a list where a vacancy can have one or multiple departments*

## QA

1. go to https://canonical-com-428.demos.haus/careers/all and make sure that the page is not broken
2. Here is an example of each vacancy case:

```json
[
  {
    "jobId": 3382736,
    "jobTitle": "Chief of Staff, Global Support Services",
    "department": "techops",
    "departments": ["techops"]
  },
  {
    "jobId": 3062022,
    "jobTitle": "Cloud Engineering - open source, remote",
    "department": "techops",
    "departments": ["engineering", "techops", "operations"]
  }
]

```

#### Explation of the JSON

The first one (with the id: `3382736`) is a job that belongs to only one department and the second one (with the id: `3062022`) belongs to multiple departments.

#### Steps

For each vacancy in the list you need to :
  * go to `/careers/[department]` and make sure that the job title is in the list of jobs positions for each department in the list that it belongs to (just do Ctrl+F with the title).
  * go to `/careers/[jobId]` and make sure that the page is not broken

## Issue / Card

Fixes #352

### DIFF
--- a/webapp/greenhouse.py
+++ b/webapp/greenhouse.py
@@ -87,7 +87,7 @@ class Vacancy:
         self.employment: str = _get_metadata(job, "employment")
         self.date: str = job["updated_at"]
         self.questions: dict = job.get("questions", {})
-        self.department: str = Department(_get_metadata(job, "department"))
+        self.departments : list = list(map(lambda d: Department(d), _get_metadata(job, "departments") or []))
         self.management: str = _get_metadata(job, "management")
         self.office: str = job["offices"][0]["name"]
         self.description: str = _get_metadata(job, "description")
@@ -133,7 +133,10 @@ class Greenhouse:
         vacancies = self.get_vacancies()
 
         def department_filter(vacancy):
-            return vacancy.department.slug == department_slug
+            for department in vacancy.departments:
+                if department.slug == department_slug:
+                    return True
+            return False
 
         return list(filter(department_filter, vacancies))
 
--- a/webapp/greenhouse.py
+++ b/webapp/greenhouse.py
@@ -87,7 +87,12 @@ class Vacancy:
         self.employment: str = _get_metadata(job, "employment")
         self.date: str = job["updated_at"]
         self.questions: dict = job.get("questions", {})
-        self.departments : list = list(map(lambda d: Department(d), _get_metadata(job, "departments") or []))
+        self.departments: list = list(
+            map(
+                lambda d: Department(d),
+                _get_metadata(job, "departments") or [],
+            )
+        )
         self.management: str = _get_metadata(job, "management")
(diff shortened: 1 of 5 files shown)


pmahnke:
Please review, but sounds like we need a greenhouse data cleanup to allow this to work correctly.

webapp/greenhouse.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, job: dict):
         self.employment: str = _get_metadata(job, "employment")
         self.date: str = job["updated_at"]
         self.questions: dict = job.get("questions", {})
-        self.department: str = Department(_get_metadata(job, "department"))
+        self.departments : list = list(map(lambda d: Department(d), _get_metadata(job, "departments") or []))
```
nottrobin, line 90: Stray space here I think

goulinkh: Is it better now @nottrobin ?

carkod:
@goulinkh you should be able to fix Python lint errors by just running `dotrun format-python` see more in [package.json](https://github.com/canonical-web-and-design/canonical.com/blob/master/package.json#L8)

webapp/greenhouse.py
nottrobin: The space, yes. But the lint check still fails. Are you able to run `black --line-length 79"? If it's still not passing maybe @jkfran could help you out?

goulinkh:
Thanks @carkod, it should be better now,  I thought that `dotrun lint-python` would do the fixing as well and in the `/run` file there is no mention for the command `format-python`

nottrobin (approved):
Solid work thanks @goulinkh

nottrobin:
I didn't QA though. But I think @pmahnke said he did?

goulinkh:
Not yet, there are some vacancies that needs to be updated in greenhouse before doing the QA and the merge I think

pmahnke:
@goulinkh I think I fixed the missing data, can you easily check?

goulinkh:
@pmahnke Thank you for the update, the missing values are fixed! I just checked the returned json and also in the demo version the total count of vacancies is the same as the prod one

PR #454, opened 2021-12-05
Add application page for candidates

tbille:
## Done

- This is a page dedicated to applications for candidates that apply to jobs. This link will be shared in the emails exchanged between the recruiter and the candidate
- A candidate can have multiple pages if the candidate applies to multiple jobs
- This page will let the candidate know what is the progress of his/her application
- The page shows a list of useful information like:
  - Job link
  - Next interview (with a link to convert the time to your own timezone)
  - Previous interviews
  - What stage is the application (WIP)
  - Some useful links about Canonical (WIP)
  - A form to delete the candidate application (WIP)
- The page is not indexed 
- The link is made with `first_name`, `candidate_id` and `application_id`. This is made to avoid someone doing a bruteforce to find all those links: `/careers/application/SUPER-ENCRYPTED-LINK`
- Only the first name of the candidate is displayed on purpose in case someone finds the link 

Here is a copydocs with the proposed content: https://docs.google.com/document/d/1R3-aIHtdrs0eD0qFI8LiUPOf-fVY6IoxAsdjpd-2WO4/edit#heading=h.4ir3ufdci83f

## QA

- Test those URLs (to avoid putting any public link the URLs are in the pastebin) https://pastebin.canonical.com/p/f49wrH9mHM/

# TODO

- [x] A cool design
- [x]  What stage is the application 
- [x] A form to delete the candidate application (Can be done in another PR)
- [x] Some useful links about Canonical
- [x] Update link to be: a unique UUID, this will come with a script that should update all the candidates to add the extra variable in greenhouse.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ canonicalwebteam.templatefinder==1.0.0
 markdown==3.3.3
 python-slugify==4.0.1
 vcrpy-unittest==0.1.7
+cryptography
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ canonicalwebteam.templatefinder==1.0.0
 markdown==3.3.3
 python-slugify==4.0.1
 vcrpy-unittest==0.1.7
-cryptography
+cryptography==36.0.1
--- a/webapp/application.py
+++ b/webapp/application.py
@@ -92,7 +92,6 @@ def application_page(token):
     if not candidate["first_name"].lower() == name:
         flask.abort(404)
 
-    print(application["current_stage"])
     if application["current_stage"]:
         application["stage_progress"] = stage_progress(
             application["current_stage"]["name"]
--- a/webapp/application.py
+++ b/webapp/application.py
@@ -117,19 +117,3 @@ def application_page(token):
         job=job,
         hiring_lead=hiring_lead,
     )
-
-
-@application.route(
-    "/<int:candidate_id>-<int:application_id>/delete", methods=["DELETE"]
-)
-def delete_application(candidate_id, application_id):
-    application = harvest.get_application(application_id)
-    if (
-        "candidate_id" not in application
-        or application["candidate_id"] != candidate_id
-    ):
-        flask.abort(404)
-
-    # CODE TO DELETE APPLICATION
-
-    return flask.render_template("application.html", application=application)
--- a/webapp/application.py
+++ b/webapp/application.py
@@ -17,7 +17,7 @@ application = flask.Blueprint(
 
 session = talisker.requests.get_session()
 harvest = Harvest(session=session, api_key=os.environ.get("HARVEST_API_KEY"))
-cipher = Cipher(os.environ.get("SECRET_KEY"))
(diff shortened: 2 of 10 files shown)


webapp/utils/cipher.py
```diff
@@ -0,0 +1,26 @@
...
+import hashlib
+
+from cryptography.fernet import Fernet, InvalidToken
+
+
+class Cipher:
+    def __init__(self, secret):
+        self.cipher_suite = Fernet(
+            base64.urlsafe_b64encode(
+                hashlib.sha256(secret.encode("utf-8")).digest()
```
nottrobin, line 11: Why do we need to hash this?

goulinkh: Because the secret key length must be 32 characters
So I hash to make sure it respects the length

requirements.txt
```diff
@@ -5,3 +5,4 @@ canonicalwebteam.templatefinder==1.0.0
 markdown==3.3.3
 python-slugify==4.0.1
 vcrpy-unittest==0.1.7
+cryptography
```
tbille, line 8: Pin dependency.

tbille:
Next steps:

- [ ] Follow up this solution with infrastructure guild (CC @nottrobin any opinions on the cryptographic algorithm used?)
- [ ] Add custom on greenhouse for candidates: CANONICAL_URL_HASH (name can be changed :) )
- [ ] Write webhook for all new applications to add the field automatically.  I think this webhook should on hiring.canonical.com. This means that we should use a common key to encrypt (so not the `SECRET_KEY`)
- [ ] Write a temporary script that updates ALL current applications to have this field filled.

webapp/utils/cipher.py
nottrobin: Yeah, but we can just make sure the secret is 32 bytes long can't we? Here you're adding processing overhead to every request. I agree this is hopefully insignificant but it's still not good to do if it can be avoided.

Secrets are always something that needs to be very specific. It's not like people can just expect to choose a random string. So expecting the secret to be stored at 32 bytes long doesn't seem to be particularly unreasonable to me...

goulinkh: 32 bytes long secret is a requirement of the cryptography library, having a secret that doen't match the requirement throw an errror. This is calculated only once at the moment we instanciate the class.

I don't mind removing this line, the problem is that we will have to update all the applications encrypted token with a new one which @tbille spent a few hours on last time!

marinacastejon (approved):
Nicely done.

webapp/application.py
```diff
@@ -0,0 +1,129 @@
...
+        "candidate_id" not in application
+        or application["candidate_id"] != candidate_id
+    ):
+        flask.abort(404)
+
+    candidate = harvest.get_candidate(application["candidate_id"])
+    if not candidate["first_name"].lower() == name:
+        flask.abort(404)
+
+    print(application["current_stage"])
```
andesol, line 95: :eyes:

webapp/utils/cipher.py
nottrobin: *sigh* fine. It's merged now anyway.

PR #560, opened 2022-06-09
Remove the full-time option on job filters

anthonydillon:
## Done
Remove the full-time option on job filters

## QA
- View the site locally in your web browser at: http://0.0.0.0:8002/career/all
- Check there is no "Full time" option in the FIlter by


minkyngkm (approved):
LGTM

PR #574, opened 2022-07-14
Add search input with JS fuzzy match

minkyngkm:
## Done
- Install `fuse.js`
- Enable search

## QA

- 

## Issue / Card

Fixes #


minkyngkm:
There might be still things to do but could you please take a quick look and give me feedback @sowasred2012?

anthonydillon:
The finance page is 500'ing

marinacastejon:
Hi @minkyngkm 
I still find the same issue as yesterday. When I search for bootstack, I get among the results [this job](https://canonical-com-574.demos.haus/careers/3271625) that doesn't have Bootstack anywhere in the job page

anthonydillon:
The auto complete is looking great but when you press enter of a term with no results the resulting page shows all the jobs. Ignore if resolved in another PR.

minkyngkm:
@anthonydillon Yes! Scott and I talked about this issue and it will be fixed with another PR! Thanks Ant!

marinacastejon:
Hi! I took another look at the component. I looks great except for one small thing. We only show up front 3 jobs and the box fits so perfectly that it is difficult to guess you can scroll down. 
image
I have a suggestion to make scrolling down more discoverable. When the user hovers over one of jobs suggested, we could display the scroll bar. The current behaviour is that the scrollbar appears when the candidate hovers **and** starts scrolling. Other than that, looks good to me, thanks!

anthonydillon:
Thanks @minkyngkm much better now. One small thing is that if you search for something like "Fast" you get no results. We should say no results for that but there is no way to clear the search and return to the main list as the search input is required. I suggest adding a clear filter link to the search area that clears the query.

minkyngkm:
Thanks @marinacastejon for your feedback! I made the scroll more visible.(some things like color and padding might need to be modified because it is a custom scroll).  Could you take a look again please?  🙏🏻

Thanks @anthonydillon for your comment! I will make sure to fix what you pointed out through this [PR](https://github.com/canonical-web-and-design/canonical.com/pull/575) Thanks! 👍

static/sass/_pattern_search-panel.scss
```diff
@@ -0,0 +1,49 @@
...
+    margin-top: -1.2rem;
+    position : relative;
+
+    .p-search-panel__container{
+      background: #fff;
+      border:  1px solid #000;
+      border-top : none;
+      max-height: 15rem;
+      overflow-y: scroll;
+      &::-webkit-scrollbar-track {
```
sowasred2012, line 15:
```suggestion

      &::-webkit-scrollbar-track {
```

static/sass/_pattern_search-panel.scss
```diff
@@ -0,0 +1,49 @@
...
+        border-radius: 100px;
+        box-shadow: inset 2px 2px 5px 0 rgba(#fff, 0.5);
+        min-height: 100px;
+        width: 0.8rem;
+      }
+      
+      position: absolute;
+      width: 100%;
+      z-index: 34;
+      .p-list__item {
```
sowasred2012, line 34:
```suggestion

      .p-list__item {
```

templates/base_index.html
```diff
@@ -78,6 +78,7 @@
 
     <script src="{{ versioned_static('js/navigation.js') }}" defer></script>
     <script src="https://assets.ubuntu.com/v1/703e23c9-lazysizes+noscript+native-loading.5.1.2.min.js" defer></script>
+    <script src="https://cdn.jsdelivr.net/npm/fuse.js@6.6.2"></script>
```
sowasred2012, line 81: Fuse is installed as a dependency in package.json, but then we link to a CDN hosted version of the script - could we add fuse to the `build-js` step, and link to that script instead?

templates/careers/_search.html
```diff
@@ -0,0 +1,108 @@
...
+  
+      const title = document.createElement('div')
+      title.setAttribute('class', 'u-truncate');
+      const anchor = document.createElement('a')
+      anchor.setAttribute('class', 'p-link--soft u-no-padding--top u-no-margin--bottom');
+      anchor.setAttribute('href', `/careers/${job.id}`)
+      title.textContent = `${job.title}`
+      // const description = document.createElement('p')
+      // description.setAttribute('class', 'u-sv1');
+      // description.textContent = job.description
```
sowasred2012, line 58: Commented out code

templates/base_index.html
sowasred2012: Also, does this need to be in the base template? Is there a careers specific template this could be moved to? Careers pages are the only ones that need to load this script.

templates/partial/_careers-available-roles.html
```diff
@@ -1,10 +1,7 @@
-{% if vacancies_json %}
-<script>
-  const vacancies = {{ vacancies_json | tojson }};
-</script>
-{% endif %}
-
 {% if vacancies %}
+<div>
+  {% include "careers/_search.html" %}
+</div>
```
sowasred2012, line 4: Is the wrapping div necessary? Could it be moved to the partial if it is?

sowasred2012 (changes requested):
Code is looking good, there's just some house keeping to do.

sowasred2012 (approved):
LGTM!

PR #575, opened 2022-07-18
Display search result

minkyngkm:
## Done

- Display search resultㄴ after pressing enter.
- Filter and sort search results. 

## QA

- Go to  https://canonical-com-575.demos.haus/careers/all
- e.g. Search "cloud" and press enter
- Check the job list displays properly 
- Check filter and sort works with the search result

## Issue / Card

Fixes #[5701](https://github.com/canonical-web-and-design/web-squad/issues/5701)

### DIFF
--- a/templates/careers/base.html
+++ b/templates/careers/base.html
@@ -2,6 +2,7 @@
 
 {% block extra_metatags %}
   <link rel="alternate" type="application/rss+xml"  href="/careers/feed" title="All job roles | Canonical Careers">
+  <script src="{{ versioned_static('js/modules/fuse/fuse.js') }}"></script>
 {% endblock %}
 
 {% block hero %}{% endblock %}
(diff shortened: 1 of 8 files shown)


sowasred2012 (changes requested):
This looks good, but doesn't address a point @anthonydillon made in the previous PR - if there are no results, we should inform the user, like we do on live.

See:
https://canonical.com/careers/finance?filter=home-based&location=all#available-roles
vs
https://canonical-com-575.demos.haus/careers/all?search=fast

minkyngkm:
Thanks @sowasred2012. I forgot about that. It's fixed now 👍

templates/careers/base.html
```diff
@@ -2,7 +2,6 @@
 
 {% block extra_metatags %}
   <link rel="alternate" type="application/rss+xml"  href="/careers/feed" title="All job roles | Canonical Careers">
-  <script src="{{ versioned_static('js/modules/fuse/fuse.js') }}"></script>
```
sowasred2012, line 5: Not sure I understand why this was moved to base_index, won't that mean this script will be loaded on pages that don't use fuse?

sowasred2012 (approved):
Nice! LGTM

PR #589, opened 2022-07-22
Simplify the navigation across the site

anthonydillon:
## Done
- Removed the navigation dropdowns
- Made the navigation consistent across the site
- Improved the secondary styling in the partners section
- Removed tons of code

## QA
- Use the navigation and see its more consistent
- Go to /partners and see the secondary nav looks better but the same

## Issue / Card
Fixes https://github.com/canonical-web-and-design/canonical.com/issues/124


davegoddard42:
@anthonydillon Overall, looks good, it's generally a big improvement on what's there. 

There's issue here from a UX POV. 

The current operation on the homepage of the nav scrolls the user to the correct section of the page. The nav is sticky. By removing the sticky nav operation, when a user navigates from Careers or Partners to the homepage, they lose the nav when landing on the prescrolled homepage. 

This makes it difficult for the user to know where they are and how to navigate to the next place - it's quite disorientating, especially as both Careers and Partners have a menu at the top when the user lands on them. Given that much of the traffic comes to this site from people searching for job, there's a good chance they did not land on the homepage as an entry point but something else. On clicking Products, they wouldn't be aware they've been redirected to the homepage, or that the menu is at the very top of the page. 

The navigation may now work the same across all pages but consistency in this instance has gone down. It's more consistent that there is a nav bar at the top when a user selects "Products" as that's what a user would expect.

**Solutions**:
- Bring back sticky nav on the homepage. Don't include the "highlight" of the menu item as the user scrolls. All other operation the same as in the demo.
- Roll out the new Product page. We have a new product page featuring all products designed. There's no reason this can't be launched using the current styling. The menu will link to this new page. The homepage would need a link added to the products section also.

anthonydillon:
I went for the simpler task for now as the product search would take a little while to work out. Ready for another look @davegoddard42

davegoddard42:
@anthonydillon - nice, that works much better. Once we've got the product page up we can get what you originally did back up and running. 

Only issue I can see is that on mobile view, the "Menu" link has disappeared - what's strange is clicking that area does open the menu, but the word has vanished! Looking closely I think it's the text colour is very dark - it's just about visible. 

**Demo:**
image

**What's live:**
image

Once more thing I've noticed - on the "Partner" pages the secondary menu on mobile is wrapped to a crazy degree - this is actually what's live so it's out of scope of this issue. I'll take a look at this though and propose a simple solution.

image

davegoddard42:
@anthonydillon - nice, that works much better. Once we've got the product page up we can get what you originally did back up and running. 

Only issue I can see is that on mobile view, the "Menu" link has disappeared - what's strange is clicking that area does open the menu, but the word has vanished! Looking closely I think it's the text colour is very dark - it's just about visible. 

**Demo:**
image

**What's live:**
image

Once more thing I've noticed - on the "Partner" pages the secondary menu on mobile is wrapped to a crazy degree - this is actually what's live so it's out of scope of this issue. I'll take a look at this though and propose a simple solution.

image

davegoddard42:
Issue opened [here](https://github.com/canonical-web-and-design/canonical.com/issues/598) to deal with Partners pages menu

anthonydillon:
@davegoddard42 fixed the mobile navigation. Thanks.

davegoddard42:
Thanks @anthonydillon - looks great

mtruj013 (approved):
LGTM, just have a linting error

Issue #581, opened 2022-07-18
[c.c application page] Enable sending email with a link triggering application withdrawal


minkyngkm:
Reach out to Goulin or Joao and discuss

PR #593, opened 2022-07-26
Send withdrawal confirmation email

minkyngkm:
## Done

- Send withdrawal confirmation email with a link 

## QA

- Go to https://canonical-com-593.demos.haus/careers/application/token (Please ping me for token) 
- Click application withdrawal button.
- Type `min.kim@canoinical.com` to send confirmation email. (Also, try any other email to check it's not sending email when the email doesn't match)
- Check the printed email in your terminal.
- Copy the link triggering application withdrawal.
- Check the printed result in the terminal. 
- [Spec](https://docs.google.com/document/d/1aENPmLdPjoRtMgmmGX2x4iewevRv905OdkKgqqm-sHk/edit)
- [Miro](https://miro.com/app/board/uXjVO9E_QFA=/?share_link_id=997105033759) for UX
- [Figma](https://www.figma.com/file/KQSCmTqOjY4gfKvJPyvzj2/Canonical.com-latest-only?node-id=4563%3A8766) for design

## Issue / Card

Fixes [#583](https://github.com/canonical/canonical.com/issues/583), [#581](https://github.com/canonical/canonical.com/issues/581), #596, [#585](https://github.com/canonical/canonical.com/issues/585)


webapp/application.py
```diff
@@ -125,4 +131,114 @@ def application_page(token):
...
+):
+    # import smtplib
+    try:
+        msg = EmailMessage()
+        msg["Subject"] = subject
+        msg["From"] = from_email
+        msg["To"] = ", ".join(to_email)
+        msg.set_content(message)
+
+        server = smtplib.SMTP(server, 587)
```
nottrobin, line 237: shouldn't overload the `server` variable. Rename one of the two.

webapp/application.py
```diff
@@ -125,4 +131,114 @@ def application_page(token):
...
+        server.starttls()
+        server.login(from_email, "")  # user & password
```
nottrobin, line 239: Isn't a password required? How is it going to get a password? Or does smtp.gmail.com literally let you send without one?

Presumably the from email isn't going to be min.kim. What's the ultimate email going to be? And will it actually work with the gmail SMTP server?

sowasred2012:
Blocked by IS: https://portal.admin.canonical.com/152319

webapp/application.py
```diff
@@ -125,4 +131,178 @@ def application_page(token):
...
+def application_withdrawal(token):
+    payload = verification_token_cipher.decrypt(token)
+    try:
+        payload = json.loads(payload)
+        email = payload.get("email")
+        withdrawal_reason = payload.get("withdrawal_reason", "")
+        candidate_id = payload.get("candidate_id")
+        application_id = payload.get("application_id")
+    except (ValueError, TypeError):
+        flask.abort(404)
```
nottrobin, line 148: This try catch block it too greedy. It's gonna be pretty misleading to mask all the potential errors here with a 404 page.

Why wouldn't you want to inform the user / developers about `ValueError`s and `TypeError`s anyway?

webapp/application.py
```diff
@@ -125,4 +131,178 @@ def application_page(token):
         application=application,
         job=job,
         hiring_lead=hiring_lead,
+        token=token,
+    )
+
+
+@application.route("/withdraw/<string:token>")
+def application_withdrawal(token):
+    payload = verification_token_cipher.decrypt(token)
```
nottrobin, line 140: I would imagine that failing to decrypt this cipher must be fairly likely. Should we be adding a `try` block here so we can display a more friendly error page than just a plain 500 when a there's a problem with the token?

webapp/application.py
```diff
@@ -18,6 +22,8 @@
 session = talisker.requests.get_session()
 harvest = Harvest(session=session, api_key=os.environ.get("HARVEST_API_KEY"))
 cipher = Cipher(os.environ.get("APPLICATION_CRYPTO_SECRET_KEY"))
+verification_token_cipher = Cipher(os.environ.get("SECRET_KEY"))
```
nottrobin, line 25: I'd like two understand the distinct things these two different ciphers are doing. I'll try to find time to delve into the code more to look into it, unless you're able to give me a quick summary.

webapp/application.py
```diff
@@ -125,4 +131,178 @@ def application_page(token):
...
+    # call the Harvest API to reject the application
+
+    rejection_reason_id = get_reason_id(withdrawal_reason)
+    notes = withdrawal_reason
+
+    harvest.reject_application(
+        application_id, candidate_id, rejection_reason_id, notes
+    )
+
+    print(application_id, candidate_id, rejection_reason_id, notes)
```
nottrobin, line 172: If it's important that we track these rejections, could we submit them to Sentry rather than the logs? In the logs they'll get immediately buried, and will be deleted after about 4 days.

webapp/application.py
```diff
@@ -125,4 +131,178 @@ def application_page(token):
...
+        or application["candidate_id"] != candidate_id
+    ):
+        flask.abort(404)
+
+    print(candidate["first_name"], withdrawal_reason)
+
+    # call the Harvest API to reject the application
+
+    rejection_reason_id = get_reason_id(withdrawal_reason)
+    notes = withdrawal_reason
```
nottrobin, line 166: Why are you renaming this variable? I would pick one or the other as the best description of the variable and stick to it consistently throughout the code.

webapp/application.py
```diff
@@ -125,4 +131,178 @@ def application_page(token):
...
+def _post_api(url, headers, payload={}):
+    return session.post(
+        url,
+        auth=(harvest, ""),
+        data=json.dumps(payload),
+        headers=headers,
     )
+
+
+def get_reason_id(withdrawal_reason):
```
nottrobin, line 296: I'm assuming we're doing this because we need to send this ID to Harvest, but Greenhouse doesn't provide any way to get the ID. Is that right? If so, this might be the best we can do, but it's a shame because this solution is pretty brittle. I doubt they consider those exact strings to be part of their API contract, so they could easily change in which case these IDs would no longer match up.

nottrobin (changes requested):
Some comments inline about a few code suggestions.

webapp/application.py
andesol: When rejecting an application you may want (optionally) to provide the rejection reason, and that's done passing the id, that's right. But this function seems to be mapping the option selected in the front-end with the id, so this string could be anything as long as it matches the `value` attribute in the form. It doesn't need to be the specific reason as defined in GH. A `dict` mapping rejection reason with id for easy lookup would do the trick, IMHO.

nottrobin: Sorry, I didn't quite follow that. Are you suggesting this function should stay the same? Or change?

andesol: Sorry, I just meant that these strings were defined [here](https://github.com/canonical/canonical.com/pull/593/files#diff-3ec962bc637c818d3d11084bdbbd92ef5e84c768ef44910b13040681554435fcR14), so it won't break if somebody renames it in GH. About the function, it can stay the same, but I suggested that having a dict may be easier than an if statement. @nottrobin 
```py
rejection_reasons_ids = {
   "reason 1": 1111,
   "reason 2": 2222
   ...
}
```

nottrobin: Sorry so these strings come through from our own form? If that's the case then we definitely shouldn't be using whole sentences like this as our `name` (basically the ID). If we do that, some future developer is clearly going to think those values can just be reworded without impact.

I suppose maybe this should be a dictionary of IDs to descriptions that we define at a more global level and pass through to the template? That would be much safer.

webapp/application.py
```diff
@@ -126,6 +134,7 @@ def application_page(token):
     return flask.render_template(
         "applications/application.html",
         title="You application",
+        withdrawal_reasons=withdrawal_reasons,
```
minkyngkm, line 137: @nottrobin I keep getting this error from here. Do you know why by any chance? 

image

minkyngkm: sorry, fixed! :)

carkod:
@nottrobin can you give a code review

webapp/application.py
```diff
@@ -18,7 +18,7 @@
     "27992": "I've decided to stay with my current employer",
     "36714": "I cannot complete the assessment",
     "35818": "The position isn't a good fit",
-    "33": "Other"
+    "33": "Other",
```
minkyngkm, line 21: This comma is mistakenly added. I will remove it.

marinacastejon:
Hi @minkyngkm 
Thank you for this. 
A few comments: 
- The send email button should not be red, since it is not a definite or irreversible action.
- The width of the paragraph in the modal ("We will send you an email with...") seems too narrow, I think it should be full width? (unless the Vanilla guidelines say otherwise)
- Not sure what is the guideline for the field width. @lyubomir-popov do you want to give some inputs? 
- From what I see in Vanilla, there should be more padding between the modal title and the divider line. 
Thank you!

lyubomir-popov:
I get the following when I visit the url: 
image

Do I need special setup/permissions?

webapp/application.py
nottrobin: oh sorry. I imagine you needed `.items()` or something?

minkyngkm: yep I added `.item()`

webapp/application.py
nottrobin: putting a comma at the end of a list of dictionary items, or arguments, or whatever doesn't matter in Python. You can either add the trailing comma or not. Adding it makes it easier to reorder the items without causing a syntax error. Generally we should just let Black take care of this, I *think* bash tends to add these commas by default. But it really doesn't matter which we do.

marinacastejon:
@minkyngkm Lyubo did the visual design for the mock-up. Maybe you want to create a separate issue for this, as you want. [Figma mock-up](https://www.figma.com/file/KQSCmTqOjY4gfKvJPyvzj2/Canonical.com-latest-only?node-id=5742%3A10933).

webapp/application.py
```diff
@@ -121,8 +134,153 @@ def application_page(token):
...
+    # import smtplib
+    msg = EmailMessage()
+    msg["Subject"] = subject
+    msg["From"] = smtp_sender_address
+    msg["To"] = ", ".join(to_email)
+    msg.set_content(message, subtype="html")
+
+    server = SMTP(smtp_server)
+    # check if smtp_server exists as it would not exist on development
+    if smtp_server:
```
nottrobin, line 277: So what you're saying here is that - in local development we don't expect the SMTP server to be configured, so we won't send an actual email.

I'm worried that this will easily lead to us completely failing to realise that our SMTP server is broken and emails potentially aren't being sent.

Could we do this through a different mechanism? I'm not exactly sure what's best. Probably, this `send_mail` function should error as normal, but then at line 238 we should catch the error *only if* we're in debug mode. We could even display the email on the resulting page or something for easier debugging?

Or we could just let it error and expect anyone who really wants to test this out properly just to configure the SMTP server. Is that super unreasonable?

minkyngkm: yep @jpmartinspt and I tried to make it work the other day but it's blocked to send an actual email through SMTP server in local development for some reason... what you think @jpmartinspt?

jpmartinspt: I'm almost sure we are not able to connect to the SMTP server from our machines, so I suggested we just print to message to the console when in development. I would not even try to connect when in development as it will likely just stall for as long as the connection timeout is set for. We can always QA the e-mail is actually being sent by using the demo(and maybe an integration test as part of the PR workflow) and let eventual staging/production errors with the sending land in Sentry.

nottrobin: Ok. But let's:

a) do it explicitly based on debug mode (or something similar) rather than the existence of a variable
b) put a message in the actual success page to say that the email wasn't actually sent

webapp/application.py
```diff
@@ -119,10 +132,172 @@ def application_page(token):
...
+    candidate_email = parseaddr(candidate["email_addresses"][0]["value"])[1]
+
+    if flask.request.form["withdrawal-reason-other"]:
+        withdrawal_reason = flask.request.form["withdrawal-reason-other"]
+    else:
+        withdrawal_reason = flask.request.form["withdrawal-reason"]
+
+    # Reject if user typed the wrong email
+    if candidate_email != email:
+        return redirect(flask.request.referrer + "#wrong-email")
```
minkyngkm, line 220: @nottrobin Would it be better to show a new page as well when a wrong email is typed? 
This works but one issue is the notification message is hidden by navigation even though the hash url points the matching id section. Would there be a better way for this?

minkyngkm:
Thanks @nottrobin. Looking great! I just left one comment about the wrong email notification. Feel free to let me know what you think! :)

webapp/application.py
nottrobin: Yes this is why I wasn't crazy about the notification solution. It feels like it can easily go wrong and the message doesn't get through.

But I'm gonna fix the other thing first and then think about this. I'll add a comment at the bottom about my conversation with Marina.

nottrobin:
Okay so I spoke to @marinacastejon about the solution for the withdrawal requested message.

Marina made the point that a dedicated "withdrawal requested" page could easily mislead the user - if they don't read it properly (as users are wont to do) - into mistakenly thinking the application has in fact already been withdrawn. They might not notice that there's a step 2. This is the reason for the original design.

This makes very good sense to me. By returning the user to the application page with a notification, we make it clearer that the application is still active for now.

I still don't like the redirect flow we had before. So instead I'm going to try to perform the POST at the original application URL, and serve the application directly back in response rather than redirecting.

I'll update this PR with this soon so people can see what they think.

nottrobin:
Okay I've updated the code in a way that I think addresses @marinacastejon's concerns. The experience now looks very similar to the original solution. I've just changed the code so it's no longer doing a redirect. And I've removed the `#` targeting, because it seemed to me to break the experience more than help - now it just returns you to the top of the page which should mean you can see the notification right up front anyway.

@minkyngkm could you please provide Marina with a token to test the demo so she can approve the UX?

If that's all ok, the code now looks good to me. I think it's now on you @minkyngkm to give my code and QA the +1.

webapp/application.py
```diff
@@ -1,13 +1,25 @@
...
 import flask
 import talisker.requests
+from dateutil.parser import parse
 
 from webapp.greenhouse import Harvest
 from webapp.utils.cipher import Cipher
 
+
+withdrawal_reasons = {
+    "": "No reason given",
```
minkyngkm, line 16: @marinacastejon is the withdrawal reason optional?

minkyngkm:
Thanks @nottrobin. It looks great!!👍 I just made some updates for the notification as there were some inline styles added.

minkyngkm:
@lyubomir-popov I will remove the `Review: design needed` label as design will be implemented by another PR! :)

marinacastejon:
Hi @minkyngkm , 
A few comments as discussed earlier:
- The reasons in the dropdown should follow what HR specified ([see copy doc](https://docs.google.com/document/d/1RZXj-LH_7Eg-pWJR9tE_V3D5CSDzCzyhyn0WoHLZwpQ/edit#bookmark=id.s451xxp23xqm))
- Email address doesn't have ":" and "Reason you wish to withdraw your application:" does. In the Vanilla examples, there are no ":". Could you please remove it then? I have updated the copydoc ;)
- When the email is correct, the withdrawal application section changes, there is no button and the copy changes as per [copy doc](https://docs.google.com/document/d/1RZXj-LH_7Eg-pWJR9tE_V3D5CSDzCzyhyn0WoHLZwpQ/edit#bookmark=id.xk8ddq7f58t1). 
- The error message when the email doesn't match is not correct, you can find the right copy [here](https://docs.google.com/document/d/1RZXj-LH_7Eg-pWJR9tE_V3D5CSDzCzyhyn0WoHLZwpQ/edit#bookmark=id.oup72kalp809).
- In this same scenario of email not matching, the text in the section Application withdrawal is also updated, there shouldn't be a button and the error message is the same as the top notification.

webapp/application.py
marinacastejon: no, both the email address and the withdrawal reason are mandatory @minkyngkm Thanks

minkyngkm:
Thanks @marinacastejon for your feedback! Everything you mentioned has been addressed but I had a question for the last one. When the email doesn't match, the notification saying "Please try again using the email you used in the application." will appear at the top, but if the button disappears, how could candidates try again?

nottrobin (approved):
This is looking great. Thanks for all the hard work and patience. I'm going to approve so this can be merged as soon as it has the UX +1.

marinacastejon:
Hi @minkyngkm , 
You are totally right, thanks for spotting my mistake. So, in the case the email doesn't match, the copy in the withdrawal application section should match the top notification but the button should still be there (label doesn't change). 
Thanks!

PR #605, opened 2022-08-04
/careers and /careers/all redesign

mtruj013:
## Done

- Applied redesign as per [figma mock-up](https://www.figma.com/file/KQSCmTqOjY4gfKvJPyvzj2/Canonical.com-latest-only?node-id=5339%3A10871)
- Updated copy as per [copy doc](https://docs.google.com/document/d/1o8ind3Zav-kwrUDWgiMJnkw2QLpniKgdlWrw9byIvlE/edit#) 

## QA

- View the site locally in your web browser at: https://canonical-com-605.demos.haus/careers/ (landing page) and https://canonical-com-605.demos.haus/careers/all (roles page)
- Check against design and copy doc and see that everything matches
- Test search functionality
- Test filters and pagination 
- Test buttons

## Issue / Card

Fixes https://github.com/canonical/web-squad/issues/5783


marinacastejon:
Also, please check heading structure follows the comments in the [copy doc](https://docs.google.com/document/d/1o8ind3Zav-kwrUDWgiMJnkw2QLpniKgdlWrw9byIvlE/edit#).

marinacastejon:
Hi  @mtruj013 ,
The button label should be "Search roles"
Other than that it looks good to me. As discussed, for now please don't remove the links and sections until we don't know if the content will be ready on time. 
Thanks!

lyubomir-popov:
- [ ] please match the headings to the mission heading on the index page (they are still a little different):

before: image

after:
image


- [ ] please do not type actual copy in uppercase, as this is purely styling - and it may change in the future. So any hardcoded uppercase will become problematic: 

image


- [ ] please strip the padding-top/bottom from list items. We haven't done a good job at explaining this, but this padding is to ensure plain text (outside of tags like `<p>` for example) is adequately spaced. When you have elements that have their own padding/margins, this list item padding is unnecessary and adds white space where we don't need it: 
image
@bartaz should we document this somewhere? 

- [ ] can we please have the vanilla strip component included and used instead of the numerous u-sv utilities? This is preferable, because in the future we can adjust strips across the board, rather than having to hunt down hundreds of utilities and change them individually

- [ ] please use [muted text](https://vanillaframework.io/docs/examples/utilities/muted-text) instead of small text for the location. the font-size should be same as paragraphs.

cumulatively, the above changes should result in something like this (inline style to be replaced by proper import of the vanilla strip component):
```
<li class="p-list__item p-strip is-shallow u-no-padding--top" data-office="Any" data-employment="Full-time" data-location="Home based - EMEA" data-management="false" data-sector="TechOps" style="
    padding-bottom: 1.5rem;
"><hr><div class="row"><div class="col-4 "><p class=" p-heading--4"><a class="" href="/careers/4151932">Business Analyst, Commercial Systems</a></p></div><div class="col-5"><p>Analyze and document business processes used by internal operations teams teams to drive process automation and ensure consistency across operational and engineering teams.</p><p class="u-text--muted u-no-margin--bottom">Home based - EMEA</p></div></div></li>
```

static/sass/_pattern_search-form.scss
```diff
@@ -0,0 +1,51 @@
+@mixin canonical-search-form {
+  .p-form--search {
+    margin-top: 18px;
```
lyubomir-popov, line 3: why px? set the viewport to 1799px  and then increase to 1800px. YOu'll see the value of 1rem change from 16 to 18px, effectively zoomingm everything in by a factor of 18/16. Anything expressed in px won't scale though, so it's always best to express things in rem, and also use the appropriate spacing variables.

lyubomir-popov (changes requested):

static/sass/_pattern_search-form.scss
mtruj013: I think I meant to delete that line altogether honestly, I don't remember why I was playing around with the margins originally but it isn't needed.

bartaz:
> * @bartaz should we document this somewhere?

@lyubomir-popov Not sure exactly what you are referring to. I suppose it's about some unnecessary margin/padding when component is put into some kind of wrapper? I guess it's the case with any element, if it has some bottom or top margin it may have to be cleared if you put it into container with its own spacing. Documenting it would be a good start, but it will not make people do it.

Would it be possible to fix it with some first-child/last-child selector in Vanilla itself? To remove the spacing when not needed?

Anyway, feel free to open an issue in Vanilla about it and we can discuss it there.

lyubomir-popov:
@mtruj013 can we please move this over to the left as in the updated [design](https://www.figma.com/file/KQSCmTqOjY4gfKvJPyvzj2/Canonical.com-latest-only?node-id=5339%3A10871):
image

lyubomir-popov:
@bartaz I ugess the question is, do we think adding this many classes on every list item is a bit excessive:
```class="p-list__item u-no-padding--top u-no-padding--bottom"```

**Comment by bartaz** [comment]
> @bartaz I ugess the question is, do we think adding this many classes on every list item is a bit excessive: `class="p-list__item u-no-padding--top u-no-padding--bottom"` - as opposed to having a list component named something like `p-list--unpadded` for example

@lyubomir-popov on each item? I thought we are talking about top/bottom spacing of a whole least, not each item. It definitely is excesive.

Looking at how it is used on this page - I don't think it's a use case for list component.
Especially "list divided" - is meant to be a list of text items divided by a line.

image

Here it seems we are using list for layout, we are putting rows and cols in there.
Why aren't they strips? or just rows with HR between them?

**Comment by bartaz** [comment]
It looks much closer to "detailed stepped list" (just without the steps)

image

So maybe we want to add a variant like this to Vanilla? But if it can be recreated with grid layout and hrs, do we need it as a component?

**Comment by lyubomir-popov** [comment]
@bartaz We need it at a lower level. For example, the sidenav is also a ul here, and we had to remove it there as well.

**Comment by lyubomir-popov** [comment]
@bartaz I should go a step back to give somecontext - the reason there's padding in uls/ols in general is that people do `<li>some arbitrary text here</li>`. If we didn't, then the height of a list item would be theline height of root text - 24px. This is often too tight for plain text.

The problem appears when we have something like a `label`, `p`, etc - which is often the case when you have more complex content as in the sidenav or job listing elements on this page. In these cases, the extra padding simply bloats thew page.

One option to resolve this would be to strip padding from lists altogether, and provide a plain-text- specific component that padds it. This is a little less new-user friendly
The other option is to just stuff everything with long u-no-padding classes, which is excessive
A third option is to have a p-list--tight or whatever we decide to name it

**Comment by bartaz** [comment]
It seems there are some invalid nested elements in sidebar, headings appear as children of list:

image

**Comment by lyubomir-popov** [comment]
> If you use `ul` `li` without any Vanilla class names they get the styling of a "text" list. That's what you expect.

The problem is if you use a `<ul>`, you need to remove the bullets and associated margin/padding. So if you simply wanted to use a ul because that's the correct semantic element when you have repeating entries, whatever you pick you have to heavily stuff with classes, or write local css - I think this is the core of the problem. We ca ndefinitely decide we are fine with class stuffing, just wanted to raise this for discussion.

**Comment by lyubomir-popov** [comment]
> If you want some other custom list, you need to build a custom list component, better to have a custom component (even locally in a project) than throwing dozens of u-no-padding class names in HTML.

I guess the question is, should this live locally in various repos - seems like a fairly common need. Or should we have a vanilla example/component.

File: static/sass/styles.scss [hunk 2]
```diff
@@ -24,7 +25,11 @@ $p-small-lh-diff: map-get($line-heights, default-text) -
 @import "pattern_strip";
 @import "pattern_tabs";
 @import "pattern_search-panel";
+@import "pattern_search-form";
 
+@include vanilla;
+@include vf-p-strip;
```
bartaz, line 31: If you include `vanilla` you don't need to include anything else from it. It will have everything that comes in the framework.

spencerbygraves:
@mtruj013 I've noticed two small things, apologies if you've already discussed them with Lyubo. (Examples below from the build).

- On the landing page team section, there is slightly less vertical white space between the sections than on the design

image

- On the roles page, the pagination is closer to the hr in the design

image

Otherwise it looks great!

Thanks.

sowasred2012 (changes requested):
Looking good! I spotted a couple of things I think could be tweaked a little:
- [ ] When using the search input with autocomplete suggestions, the dropdown doesn't stick to the width of the input:
Screenshot from 2022-09-14 10-29-11

- [ ] When there are no more results to show, I think this button should be disabled. @marinacastejon what do you think?
Screenshot from 2022-09-14 10-30-18

mtruj013:
@sowasred2012 @marinacastejon  About disabling the "show 20 more" button, what I'd understood originally was that at the end of the list (when there are less than 20 jobs remaining) instead of the "show 20 more" button we should show the "show all" button:
Screenshot 2022-09-14 at 17 31 06

and then when that is clicked remove both buttons: 
Screenshot 2022-09-14 at 17 31 14

I've fixed the bug wherein that functionality wasn't happening, but I can totally change it to be a disabled "show 20 more" button instead if you guys prefer.

nottrobin:
A few points / questions:

- [General] The buttons for "Defining excellence" and "Our vision of remote work" don't do anything. Are these links still to be added to this PR? Or are they coming in a follow-up PR?
- [UX] Could we give users some initial prompt as to what sort of keywords exist in our taxonomy? Otherwise how are the users supposed to guess what sort of keywords to put in the search box? I've done a fair bit of looking for jobs (😱) and I would never have any idea how any given company would choose to describe the jobs I might be interested in. Am I an "engineer", a "developer", or an "architect" or maybe "consultant" or "lead"? "Front-end", "Back-end", "Full stack" or "DevOps"? I do not strongly identify as a "backend engineer" or a "devops expert" or "architect". Instead I'd look at what jobs a company has and see if I think I might fit into it. I expect I'm not alone. With this search box it's like I'm groping around in a dark room of unknown size feeling for things I might want. If you know what I mean.

mtruj013:
@nottrobin Just to reply to your first point, those pages are coming in a follow-up PR. Once approved, we're merging this one into a feature branch until those are ready and once those are also merged to said branch this will be ready to go live.

webapp/app.py
```diff
@@ -215,6 +216,39 @@ def start_career():
     return flask.render_template("/careers/start.html")
 
 
+@app.route("/careers", methods=["GET"])
```
nottrobin, line 219: (Just to let you know, it's not actually necessary to add `methods=["GET"]`, since that's the default. You can just not provide a second argument, which is what we usually do I think. But I don't mind it at all 'cos maybe there's also value in explicitly telling the developer which HTTP methods are matched. So it's up to you.)

webapp/app.py
```diff
@@ -215,6 +216,39 @@ def start_career():
...
+        vacancy.to_dict() for vacancy in context["vacancies"]
+    ]
+
+    jobs_list = context["vacancies_json"]
+
+    return flask.render_template(
+        "/careers/all.html",
+        **context,
+        jobs_list=jobs_list,
+    )
```
nottrobin, line 249: Some of this seems unnecessary.

I can't see any reference in the templates to `jobs_list`, so I don't *think* you need to pass it through to the template. Correct me if I'm wrong.

And then I don't see what's stopping you from instantiating `all_departments` and `vacancies_json` as you pass them through to the template. That way you avoid the potentially confusing `**context` syntax.

Also, is there a reason you are doing `to_dict` rather than using `json.dumps()`? I *think* it should do the same thing.

``` python3
# At the top
import json
```

Then:

```suggestion
    vacancies = greenhouse.get_vacancies()

    return flask.render_template(
        "/careers/all.html",
        all_departments=_group_by_department(greenhouse.get_vacancies()),
        vacancies=vacancies,
        vacancies_json=json.dumps(vacancies)
    )
```

nottrobin:
> [UX] Could we give users some initial prompt as to what sort of keywords exist in our taxonomy? Otherwise how are the users supposed to guess what sort of keywords to put in the search box? I've done a fair bit of looking for jobs (scream) and I would never have any idea how any given company would choose to describe the jobs I might be interested in. Am I an "engineer", a "developer", or an "architect" or maybe "consultant" or "lead"? "Front-end", "Back-end", "Full stack" or "DevOps"? I do not strongly identify as a "backend engineer" or a "devops expert" or "architect". Instead I'd look at what jobs a company has and see if I think I might fit into it. I expect I'm not alone. With this search box it's like I'm groping around in a dark room of unknown size feeling for things I might want. If you know what I mean.

Having understood this flow a bit better now, maybe what I'm saying is that I'd rather that this landing page simply had a button for "All roles/careers/jobs/vacancies/whatever" at the top. Since the current experience of the `/careers/all` neatly addresses my concern already by showing the filterable keywords on the left. I feel like the initial presentation of that search box is much more confusing than simply taking people to the next page would be.

(But then I appreciate this opens up the question of what's the difference between the top "above the fold" section and the "All departments" section lower down. I don't know the answer.)

templates/careers/_search.html
```diff
@@ -141,17 +157,22 @@
...
 {% endif %}
-<div class="js-search-jobs-form u-hide">
-  <form class="p-search-box">
-    <label class="u-off-screen" for="search">Search</label>
-    <input type="search" id="search" class="p-search-box__input js-careers__search-input" name="search" placeholder="Search" required="" autocomplete="off">
-    <button type="reset" class="p-search-box__reset"><i class="p-icon--close">Close</i></button>
-    <button type="submit" class="p-search-box__button"><i class="p-icon--search">Search</i></button>
+
+<div class="js-search-jobs-form u-hide u-align--bottom" style="display: flex;">
+  <form action="/search" class="p-form p-form--inline p-form--search">
```
nottrobin, line 162: Clicking "Search roles" takes me to https://canonical-com-605.demos.haus/search?search=, which is a 404.

webapp/app.py
```diff
@@ -215,6 +216,39 @@ def start_career():
...
+
+    return flask.render_template("/careers/index.html", **context)
+
+
+@app.route("/careers/all", methods=["GET"])
+def all_jobs():
+    context = {
+        "all_departments": _group_by_department(greenhouse.get_vacancies())
+    }
+    context["vacancies"] = greenhouse.get_vacancies()
```
nottrobin, line 237: So, we simply pass through all vacancies to the page.

What, then, is the point of only showing the first 20 and having a "show 20 more" button? Normally I would assume this is so we don't have to send all the data initially, presumably reducing the size of the payload so the initial page loads quicker, but this isn't the case here. Which means there are two further possible benefits:

- Maybe even though we're sending all the data, the page *renders* slower if we display all the roles right away. I'm sceptical about this, I don't think it's self-evident. But if this is part of the rationale here I'm happy to put it to the test.
- Maybe we think that the "show 20 more" button is a better user experience? Again, I'm sceptical. I think it's surely most convenient for the user to simply scroll down the page to see more roles (we might consider making the filter list on the left sticky), plus they can then easily use CTRL+f to search the list without having to click the button a bunch of times first. But again if I've missed some UX arguments here I'm happy to hear them.

So if I'm right about this, then the question becomes: Is it too expensive to send all roles in the initial pageload? should we only send 20 in the first load and lazy-load more of them after the fact? Or is it in fact fast enough with sending all roles in the first load that we could simply display the full list right at the start?

From playing with the demo, there doesn't seem to be any significant delay on loading that page really. It seems surprisingly quick. So I wonder if we might want to just do away with the "show 20 more" stuff completely and just display the whole list?

marinacastejon:
Hi , 
Thanks for raising these questions:

### **Keyword suggestions**
> [UX] Could we give users some initial prompt as to what sort of keywords exist in our taxonomy? Otherwise how are the users supposed to guess what sort of keywords to put in the search box? I've done a fair bit of looking for jobs (😱) and I would never have any idea how any given company would choose to describe the jobs I might be interested in. Am I an "engineer", a "developer", or an "architect" or maybe "consultant" or "lead"? "Front-end", "Back-end", "Full stack" or "DevOps"? I do not strongly identify as a "backend engineer" or a "devops expert" or "architect". Instead I'd look at what jobs a company has and see if I think I might fit into it. I expect I'm not alone. With this search box it's like I'm groping around in a dark room of unknown size feeling for things I might want. If you know what I mean.

Based on the research + benchmark + search behaviour I believe this is the optimal design. Several things to consider:
- The search behaviour shows type-ahead suggestions aka autocomplete, giving users hints of our taxonomy
- The search behaviour retrieves all jobs that match the entered keyword either in the job title or in the job description. Meaning that it is not critical that you guess the exact job title. 
- Last but not least, it'd be difficult to suggest keywords in a pool of multi-disciplinary jobs. 

**Search field in the landing page**
> Having understood this flow a bit better now, maybe what I'm saying is that I'd rather that this landing page simply had a button for "All roles/careers/jobs/vacancies/whatever" at the top. Since the current experience of the /careers/all neatly addresses my concern already by showing the filterable keywords on the left. I feel like the initial presentation of that search box is much more confusing than simply taking people to the next page would be.
>(But then I appreciate this opens up the question of what's the difference between the top "above the fold" section and the "All departments" section lower down. I don't know the answer.)

One of the most important improvements what having a search field in both the landing and the open roles pages. I understand your question is why not have only a "Search roles" button in the landing page? There are a few things:
- It is a common pattern and I don't have any reasons to believe it would be confusing to have a search field and a button. The button is visually separated from the input field, conveying that it works where you have entered a keyword or not. We also use this pattern in [Snapcraft](https://snapcraft.io/)
- Benchmark: this is quite a common standard among the career sites we benchmarked. 
- It adds the value of giving more freedom to the user: some people might want to enter a keyword from the beginning of the flow (saving them one click), other people might prefer go to the open roles pages and then use the search field and the filters. 


I'm happy to chat about any of this in a call if you prefer. As usual thanks for your thought-provoking questions.

nottrobin:
@marinacastejon Okay fair enough if you're sure about the search box then I'm happy to accept it.

The only point I'd still like to push on is about the "show 20 more" button. Unless I misunderstood, the only one of your arguments that actually *supports* this button is the following (the others are more rather tangential points that also help smooth the user's experience):

> Showing many results by default is not usable either: nobody is gonna go through 240 jobs. On top of that, showing 240 results makes the pages more difficult to navigate with a keyboard or for screen readers.

I don't understand this, maybe I missed the crux of your point. I don't understand why having 230 jobs off-screen is unusable or makes the page more difficult to navigate. They're off screen. The jobs that are most relevant are hopefully at the top of the page, on-screen. If you want more, you scroll down. If you do want to read down the list, why is it better to make you take 2 actions (scroll down + click "show 20 more" button) rather than 1 action of just scrolling down?

Of course it's quite possible that having a very long list of out-of-sight jobs down the page does hinder the user's experience in a way I don't understand. I'm not that familiar with how screen readers work. But from my limited experience I believe that the reader keeps reading down the page until the user asks it to stop, which is basically the same as scrolling. In both cases I would have thought the user is only empowered by the ability to do a full-text search on the page, and I can't see a significant way in which they are hindered by the long page as they can just ignore it.

marinacastejon:
@nottrobin I get your points but I still think this is the best option, that went through QA with UX and that was discussed in the Accessibility Guild. Plus we need to move forward with development. 
I am open to re-discussing pagination in general, since I think your points could apply to many of the places where we are using pagination so I'd open this discussion globally to best practices regarding pagination and if at the end of that conversation we need to iterate on some of the Canonical pages where we use this pattern then we can do that. 
For now let's keep this pattern as it was proposed.

templates/careers/_search.html
```diff
@@ -7,56 +7,69 @@
...
-      const title = document.createElement('h2')
-      title.setAttribute('class', 'p-heading--3 u-no-padding--top u-no-margin--bottom');
-      const anchor = document.createElement('a')
-      anchor.setAttribute('class', 'p-link--soft');
-      anchor.setAttribute('href', `/careers/${job.id}`)
-      anchor.innerHTML= `${job.title}&nbsp;&rsaquo;`
+      location.setAttribute('class', 'u-text--muted');
+      location.textContent = job.location
+
+      const title = document.createElement('p')
```
sowasred2012, line 42: I think this should be a `h3` with `p-heading--4` class (rather than giving that class to the anchor tag on line 44) - I would expect to see the job titles listed in the headings heirarchy.

sowasred2012 (commented):
One small comment in the code, and also this - when you type anything into the search input and the dropdown appears, the vertical space around it reduces. I think it might be to do with flex, I'm not sure - if you want to pair on it I'm happy to:
Peek 2022-09-21 11-01

static/sass/_pattern_search-panel.scss
```diff
@@ -3,8 +3,9 @@
   @include vf-u-text-muted;
 
   .p-search-panel {
-    margin-top: -1.2rem;
-    position : relative;
+    margin-top: 2.4rem;
```
sowasred2012, line 6: I see a tiny gap at 2.4rem:
```suggestion
    margin-top: 2.3rem;
```

lyubomir-popov: would this keep the text on the baseline grid? best to use the correct nudge variables in those calculations.

PR #609, opened 2022-08-10
update job postings to allow video embeds

sowasred2012:
## Done

This was a request from Thibaut:
- updated the job detail page to allow youtube and vimeo videos added in greenhouse to show on the page
- added logic to ensure rendered iframes come only from youtube or vimeo

## QA

- visit https://canonical-com-609.demos.haus/careers/2955155
- see that a youtube video appears in the job description

The best way to verify that the youtube and vimeo restrictions work is:
- Check out this branch
- Update [this line](https://github.com/canonical/canonical.com/pull/609/files#diff-55f12a9f17b4ba61d3e37852150b7d78b8aedbe3629bc4ce8173d4b572297a54R461) to anything other than the two domains that are now
- `dotrun` the project
- visit http://0.0.0.0:8002/careers/2955155
- see that the video does not appear

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -2,7 +2,7 @@
 import datetime
 import os
 import re
-from urllib.parse import parse_qs, urlencode
+from urllib.parse import parse_qs, urlencode, urlparse
 
 import bleach
 import flask
@@ -425,6 +425,46 @@ def slug(text):
 @app.template_filter()
 def markup(text):
     return markdown.markdown(text)
+
+
+@app.template_filter()
+def filtered_html_tags(content):
+    content = content.replace("<p>&nbsp;</p>", "")
+    allowed_tags = [
+        "iframe",
+        "h2",
+        "h3",
+        "h4",
+        "h5",
+        "h6",
+        "p",
+        "a",
+        "strong",
+        "ul",
+        "ol",
+        "li",
+        "i",
+        "em",
+        "br",
+    ]
+    allowed_attributes = {"iframe": allow_src}
+
+    return bleach.clean(
+        content,
+        tags=allowed_tags,
+        attributes=allowed_attributes,
+        strip=True,
+    )
+
+
+def allow_src(tag, name, value):
+    allowed_sources = ["www.youtube.com", "www.vimeo.com"]
+    if name in ("alt", "height", "width"):
+        return True
+    if name == "src":
+        p = urlparse(value)
+        return (not p.netloc) or p.netloc in allowed_sources
+    return False
 
 
 @app.errorhandler(502)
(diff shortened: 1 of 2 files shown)


sowasred2012:
@Lukewh @steverydz could you take a look at this? I'm in two minds about allowing iframes - on the one hand it's content coming from Greenhouse, so we could presume/be reasonably confident it's safe, but on the other hand it's allowing iframes, so could be abused.

steverydz:
Are the videos always intended to be from YouTube? If so we could check the domain maybe before allowing it?

sowasred2012:
> Are the videos always intended to be from YouTube? If so we could check the domain maybe before allowing it?

Looks like it's possible (https://bleach.readthedocs.io/en/latest/clean.html#using-functions), I'll play around with it a bit more

sowasred2012:
@steverydz I've updated with your suggestion, could you take another look?

carkod:
> @Lukewh @steverydz could you take a look at this? I'm in two minds about allowing iframes - on the one hand it's content coming from Greenhouse, so we could presume/be reasonably confident it's safe, but on the other hand it's allowing iframes, so could be abused. Would like your thoughts!

Maybe @nottrobin has some input, but we use (and everyone uses?) iframes comprehensively, I don't know any issues about potential abuses using iframe.

webapp/app.py
```diff
@@ -427,6 +427,46 @@ def markup(text):
...
+
+    return bleach.clean(
+        content,
+        tags=allowed_tags,
+        attributes=allowed_attributes,
+        strip=True,
+    )
+
+
+def allow_src(tag, name, value):
```
carkod, line 460: - `tag` is not used anywhere? 
- Does it work this way? As far as I know, Python doesn't have hoisting, the function should be defined before?

carkod:
I don't really see the point of `filtered_tags` function, I find it too restrictive

webapp/app.py
sowasred2012: Bleach passes those three arguments implicitly when functions are called this way, we just don't use the `tag` argument in the function being called. And yes, it works.

sowasred2012:
> Yes I'm primarily interested in where this input comes from. If it's from Greenhouse where only Canonical employees have access, I don't think we need to worry too much about XSS-type attacks from our own employees. If one of our employees is compromised or acting maliciously, I think we have bigger problems. So I'd agree with Carlos, we don't need to sanitise inputs so super carefully here unless there's actually a potential for input from the general public.

Yep, it's only from Greenhouse - what you've both said makes sense, just wanted a second opinion.

carkod (approved):
LGTM

PR #626, opened 2022-09-06
Application page token format

goulinkh:
## Done

- Fix a hiring lead's name and photo
- Simplify the application page token payload

## QA

- Ping me so I can send you a working demo application URL


nottrobin:
Just let me know as soon as you have another link to test with.

anthonydillon:
I have QA'ed this with a few candidates and its working well. :+1:

carkod:
@anthonydillon can you add the QA +1?

nottrobin:
@carkod we're just trying to get the demo to access the SMTP server properly so we can test this end-to-end (which is what we should be doing in the demo server). I'll give QA+1 once we've resolved that.

anthonydillon:
@nottrobin does this PR have anything to do with the email? I thought it was access to the GH data?

nottrobin:
@anthonydillon you're right, this change is only used inside `_get_application_from_token`, which is used for just the plain index page, so we could say that the fact that the index page loads is enough of a test.

However, it is also central to the whole of the withdrawal flow, is used in the `request_withdrawal` view. And, when we built that flow in https://github.com/canonical/canonical.com/pull/593, we clearly didn't test the actual sending of the email, instead skipping that step in development mode and simply echoing the contents of the email that would have been sent back out onto the page.

So I think it's prudent to check that the email flow works here. As well as this, it is important that we treat the demos as a real environment that should work exactly as production does. Now that we've discovered a place where it doesn't (the email sending hangs) it's definitely good practice to make sure we resolve that issue before simply moving on.

Unless this is holding up important work that's needed to meet a deadline? In that case we could talk definitely consider merging this anyway.

anthonydillon:
No, best to do your properly. I was just checking.

anthonydillon:
Approved this one to unblock other PRs but have created an issue to track the remaining task: https://github.com/canonical/canonical.com/issues/627

PR #629, opened 2022-09-14
Fix missing candidate object on application page

goulinkh:
## Done

- fix missing candidate object on application page

## QA

- _Please ask me for a failing application page link_
- Make sure that the page isn't broken

### DIFF
--- a/webapp/application.py
+++ b/webapp/application.py
@@ -211,11 +211,13 @@ def faq():
 
 @application.route("/<string:token>")
 def application_index(token):
+    application = _get_application_from_token(token)
     return flask.render_template(
         "careers/application/index.html",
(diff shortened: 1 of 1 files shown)


goulinkh:
> Not connected to these changes but this role seems to have two jobs which is causing the hiring lead header to be skipped.

@anthonydillon It seems to be intentionaly hidden, where [if the application is in the assessment stage](https://github.com/canonical/canonical.com/blob/main/templates/careers/application/index.html#L75) the hiring lead full profile is hidden,[ if the application stage is early stage](https://github.com/canonical/canonical.com/blob/main/templates/careers/application/index.html#L110) and forward the full profile is shown.

anthonydillon (approved):

PR #655, opened 2022-10-12
Update the heading before search for careers

anthonydillon:
## Done
Update the heading before searching for careers

## QA
- Check the code.


petesfrench (approved):
LGTM

Issue #657, opened 2022-10-13
Hiring link failing


anthonydillon:
It seems that the “We are hiring” link at the top right of the page is broken on https://ubuntu.com/

>500: Server error
Something’s gone wrong.
The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.
Try reloading the page. If the error persists, please note that it may be a [known issue](https://github.com/canonical-web-and-design/canonical.com/issues). If not, please [file a new issue](https://github.com/canonical-web-and-design/canonical.com/issues/new).

PR #658, opened 2022-10-13
Fix the /career/start path

anthonydillon:
## Done
- Update the template path in the route

## QA

- Open the demo and add `/careers/start?_ga=2.14098167.1110780800.1665671000-167831397.1628077759` to the URL
- Check it works and in production, it doesn't

## Issue / Card

Fixes https://github.com/canonical/canonical.com/issues/657


minkyngkm (approved):
(Locally QAed) LGTM

PR #668, opened 2022-10-21
Pass the gh_src URL param to the application

anthonydillon:
## Done
- Set a session storage key if the URL contains the param `gh_src`
- On the application page, read the session key and if populated sets the form field `mapped_url_token` which is [used by Greenhouse](https://developers.greenhouse.io/job-board.html#submit-an-application) to set the source of the application
- _Drive by_: Remove the search and filter JS from the apply page which was causing an error

## QA
- Open the demo and go to /careers/all
- Click on a job and inspect by the "Submit application" button that the input named `mapped_url_token` has a blank value
- Go back to the list
- Add the URL param `gh_src=campaign123` and load the page with that
- Check for errors in the console
- If none, visit a job again and see the `mapped_url_token` field is populated with "campaign123"
- Submit a test application to see it all works

### DIFF
--- a/static/js/source-inheritance.js
+++ b/static/js/source-inheritance.js
@@ -1,0 +1,22 @@
+function setSourceField() {
+  if (sessionStorage.getItem("gh_src")) {
+    const ghsrcField = document.getElementById("ghsrc");
+    const ghsrc = sessionStorage.getItem("gh_src");
+    if (ghsrc && ghsrcField) {
+      ghsrcField.value = ghsrc;
+    }
+  }
+}
+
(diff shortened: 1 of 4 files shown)


steverydz:
I've tried submitting an application a few times and each time I get `Error 400: Bad Request. Please try again!`

steverydz:
Also the `Back to list` link stays on the same page when on a job page

steverydz (approved):
LGTM 👍

anthonydillon:
@steverydz Cool done.

PR #675, opened 2022-10-30
fix broken products navigation anchor

### LINKED ISSUE #674: Bad link for Products in the nav bar
The "Products" link in the top nav bar uses an anchor with a relative URI. This causes the link to be broken when you've previously navigated to a sub-directory in the domain because the anchor link is then relative to the sub-directory instead of the root, where it is supposed to be.

dustinblack:
## Done

Simply added a slash in front of `#products` for the link so that the path is absolute to the domain root instead of relative to the current navigation location.

## Issue / Card

Fixes #674

### DIFF
--- a/templates/partial/_navigation.html
+++ b/templates/partial/_navigation.html
@@ -30,7 +30,7 @@
           </ul>
         </li>
         <li class="p-navigation__item">
-          <a href="#products" class="p-navigation__link">Products</a>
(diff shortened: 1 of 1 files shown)


sowasred2012:
Sorry, didn't see this PR! The fix has been merged and is live, again, thanks for flagging this and for taking the time to raise a PR to fix it.

PR #683, opened 2022-11-11
Fix the stepper styling

anthonydillon:
## Done
Fix the stepper styling

## QA
- Check application page stepper looks better

Before: 
image

After: 
image


minkyngkm (approved):
LGTM

PR #687, opened 2022-11-14
Remove the behavioural traits part from the Psychometric assessment s…

anthonydillon:
…ection

## Done
Simply removed the "behavioural traits" part from the Psychometric assessment section as it is now later in the process

## QA

- Check the code


carkod (approved):
LGTM

PR #695, opened 2022-11-24
Fix become a partner submission form

carkod:
## Done

Fix Become a partner submission form. Currently it's failing with nginx error

## QA

- Go to https://canonical-com-695.demos.haus/partners/become-a-partner
- Fill out the "About you" form (you can do it with "test" data)
- If submission is successful, you should be redirected to the thank-you page

Additional checks:
- Marketo should get a submission

## Issue / Card

Fixes https://github.com/canonical/ubuntu.com/issues/12287

### DIFF
--- a/templates/partners/become-a-partner.html
+++ b/templates/partners/become-a-partner.html
@@ -103,7 +103,7 @@
       </ul> #}
     </div>
     <div class="col-6 col-start-large-7">
-      <form action="https://pages.ubuntu.com/index.php/leadCapture/save" method="post" id="mktoForm_1260">
+      <form action="https://ubuntu.com/marketo/submit" method="post" id="mktoForm_1260" onsubmit="ga('send', 'Become a partner', 'Signup', 'Become a partner application');">
         <fieldset>
           <div class="p-form--container">
             <p class="p-heading--3">About you</p>
@@ -129,6 +129,17 @@
                 <label for="Phone">Phone</label>
                 <input required id="Phone" name="Phone" maxlength="255" title="" type="tel" class="mktoField mktoTelField  mktoRequired">
               </div>
+
+              {# These are honey pot fields to catch bots #}
+              <li class="u-off-screen">
+                <label class="website" for="website">Website:</label>
+                <input name="website" type="text" class="website" autocomplete="off" value="" id="website" tabindex="-1" />
+              </li>
+              <li class="u-off-screen">
+                <label class="name" for="name">Name:</label>
+                <input name="name" type="text" class="name" autocomplete="off" value="" id="name" tabindex="-1" />
+              </li>
+              {# End of honey pots #}
 
               <div class="col-3">
                 <label for="Country">Country</label>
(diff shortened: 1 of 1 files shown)


templates/partners/become-a-partner.html
```diff
@@ -429,20 +440,11 @@ <h3 class="p-muted-heading">Partnering with Canonical gives you:</h3>
...
-              <input type="hidden" name="munchkinId" class="mktoField" value="066-EOV-335">
               <input type="hidden" name="lpurl" class="mktoField" value="https://pages.ubuntu.com/become-a-partner.html?cr={creative}&amp;kw={keyword}">
-              <input type="hidden" name="cr" class="mktoField" value="">
-              <input type="hidden" name="kw" class="mktoField" value="">
-              <input type="hidden" name="q" class="mktoField" value="">
             </div>
           </div>
         </fieldset>
         <input type="hidden" name="returnURL" value="https://canonical.com/partners/thank-you" />
-        <input type="hidden" name="retURL" value="https://canonical.com/partners/thank-you" />
```
carkod, line 445: @caldav I removed these fields, they look like legacy stuff. But can you confirm that we don't need them anymore? Do we maybe need to replace it with `utm_source` fields?

templates/partners/become-a-partner.html
```diff
@@ -103,7 +103,7 @@ <h3 class="p-muted-heading">Partnering with Canonical gives you:</h3>
       </ul> #}
     </div>
     <div class="col-6 col-start-large-7">
-      <form action="https://pages.ubuntu.com/index.php/leadCapture/save" method="post" id="mktoForm_1260">
+      <form action="https://ubuntu.com/marketo/submit" method="post" id="mktoForm_1260" onsubmit="ga('send', 'Become a partner', 'Signup', 'Become a partner application');">
```
sowasred2012, line 106: It works, but should we be submitting it to a different domain still?

sowasred2012 (approved):
LGTM

templates/partners/become-a-partner.html
carkod: mm this is what we are doing at https://github.com/canonical/canonical.com/blob/9889a905623e26b2cb487f094d9b7d6e28f8c858/templates/blog/newsletter-form.html#L7

Do we need to create the endpoint in this project? It would create duplication in many places.

sowasred2012: Could we turn it into a module to use across projects? I think this site is having the same issue: https://github.com/canonical/web-squad/issues/6146

carkod: Well, I can discuss it with the infra team, but I think there could be other ideas, e.g. Joao suggested a service for discourse so this could be part of that. For now, I think the ubuntu.com one works well?

carkod: Created issue to discuss this https://github.com/canonical/web-squad/issues/6241

PR #699, opened 2022-11-25
Allow job post id to be null

### LINKED ISSUE #698: Job role id is null for some applications
## Summary
There are cases when the job role id is null. This is if the candidate doesn't apply and they are referred. 
We should not rely on this field for rendering the application page

## Process
- Open [this link](https://pastebin.canonical.com/p/qvsk2QY7Ss/) and visit the candidate
- Visit there application page in custom application fields
- See 500

andesol:
## Done

Since some applications don't have a job post id, and this is only used in the heading, I think the simplest option is use internal Greenhouse job name if the job post is not available.

## QA

- `dotrun`
-  View the site locally in your web browser at: http://0.0.0.0:8002/
- Go to the candidate page in #698 and check that it's rendered

## Issue / Card

Fixes #698

### DIFF
--- a/webapp/application.py
+++ b/webapp/application.py
@@ -40,7 +40,7 @@ def _get_application(application_id):
     application = harvest.get_application(int(application_id))
     job_post_id = application["job_post_id"]
     application["job_post"] = (
-        harvest.get_job_post(int(job_post_id)) if job_post_id else None
+        harvest.get_job_post(job_post_id) if job_post_id else None
     )
 
     # Add candidate object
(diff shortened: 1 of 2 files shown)


webapp/application.py
```diff
@@ -39,7 +39,9 @@
 def _get_application(application_id):
     application = harvest.get_application(int(application_id))
     job_post_id = application["job_post_id"]
-    application["job_post"] = harvest.get_job_post(int(job_post_id))
+    application["job_post"] = (
+        harvest.get_job_post(int(job_post_id)) if job_post_id else None
```
goulinkh, line 43: Another way of writing this is:
```py
if job_post_id:
  application["field"] = value
```
and casting to `int()` is not required.

But it's up to you, which style you prefer.

andesol: Removed the `int`, thanks!

I prefer your style. The problem is that you don't set the value explicitly to `None`, `job_post` won't exist in the application dict, and trying to access it will result in a `KeyError`. Or I think so...

goulinkh (approved):
LGTM :+1:

PR #704, opened 2022-11-28
Get the current milestone based on the historical list of stages

goulinkh:
## Done

- Instead of only getting the current milestone based on the current, this PR aims to look for the all the previous stages in case the latest stage isn't recognized by the app

## QA

- Make sure that the application page is still working for different candidates

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -1,6 +1,6 @@
 import unittest
 
-from webapp.application import _milestones_progress
+from webapp.application import _milestones_progress, _sort_stages_by_milestone
 
 all_stages = [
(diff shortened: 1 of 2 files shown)


anthonydillon:
@goulinkh I made a few small tweaks to align with randomly ordered applications. I QA'ed it across loads of jobs and all seems to be working fine now. Can you do a code review please.

anthonydillon (approved):
LGTM, just need a reverse code review

goulinkh:
The logic looks good to me, just added a few touches to optimize the code :+1:

Issue #708, opened 2022-11-29
https://canonical.com/careers/diversity returns a 500 error


basak:
*Note - filling in this form will file an issue against the canonical.com website*

## Summary

https://canonical.com/careers/diversity returns a 500 error

## Process

Browse https://canonical.com/careers/diversity

## Current and expected result

Expected: a regular web page. Actual:

> 500: Server error
> 
> Something’s gone wrong.
> 
>     The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.
> 

## Screenshot

N/A.

## Browser details

N/A

PR #710, opened 2022-11-30
Fix sorted departments

carkod:
## Done

Fix sorted departments going 500 by hiding the departments filter.
Or maybe we should hide the entire filters section?

## QA

## Issue / Card

Fixes https://github.com/canonical/canonical.com/issues/708


mtruj013 (approved):
Thanks @carkod LGTM! And yes, I think we should remove the filters from these pages. It doesn't make sense to have them as these pages are no longer just tabs on the all roles page. I'll make a follow up pr

PR #712, opened 2022-12-01
fix the case where there are multiple hold stages

goulinkh:
## Done

- Handle the case where there are multiple stages with the same name

## QA

- Make sure that the application page is working


anthonydillon (approved):
LGTM :+1:

PR #717, opened 2022-12-02
Remove get_partner_groups from homepage view

### LINKED ISSUE #196: Move partners feed client-side
Loading the homepage is quite slow 'cos it loads many partners feeds. This is not terrible as we have caching, but every now and then a user will experience a load much slower than necessary.

Really things like the partners feed should be client-side, as they're non-essential.

nottrobin:
Remove the dependency on the partners API from the index view.

Fixes #196

## QA

Go to [the demo](https://canonical-com-717.demos.haus), check the homepage works.

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -42,8 +42,7 @@ app.register_blueprint(application, url_prefix="/careers/application")
 
 @app.route("/")
 def index():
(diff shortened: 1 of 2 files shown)


jkfran (approved):
LGTM

goulinkh (approved):
LGTM

PR #723, opened 2022-12-08
Fix departments to handle new response format from API

nottrobin:
## QA

Go to /careers in the demo. Search for something. See that it works, no errors.


mtruj013 (approved):
🎉 🎉 🎉

PR #754, opened 2023-01-13
Fix featured and fast-track jobs

nottrobin:
The issue was that in the old code, we were comparing department slugs
to department names.

Much of that code was unnecessary and has now been replaced.

## QA

Go to https://canonical-com-754.demos.haus/careers/engineering, see featured jobs. Click through the other departments, check everything works (department pages code has changed significantly).


anthonydillon (approved):
Code looks good and the roles are back. The application still works.

PR #763, opened 2023-01-23
Update footer format

mtruj013:
## Done

- Updated footer as per [design](https://www.figma.com/file/KQSCmTqOjY4gfKvJPyvzj2/Canonical.com-latest-only?node-id=6990%3A16546)

## QA

- View the site locally in your web browser at: https://canonical-com-763.demos.haus/
- Scroll to the bottom of any page and see that the footer matches the proposed design
- Check that links work as expected

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-1004

### DIFF
--- a/static/sass/_pattern_footer.scss
+++ b/static/sass/_pattern_footer.scss
@@ -1,15 +1,35 @@
 @mixin canonical-p-footer {
-  .row:first-child{
-    position: relative;
+  .p-footer {
+    @extend %vf-strip;
 
-    .p-inline-list{
-      bottom: 0;
-      position: absolute;
+    background-color: #262626;
+    
+    &.is-dark {
+      color: #fff;
+    }
+  
+    .p-list__item{
+      padding-bottom: 0;
+      padding-top: 0;
+    }
+
+    a {
+      color: #3097FF;
+
+      &:visited {
+        color: #AB83D3;
+      }
     }
   }
 
-  .p-list__item{
-    padding-bottom: 0;
-    padding-top: 0;
+  @media (min-width: $breakpoint-large) {
+    .row:first-child{
+      position: relative;
+  
+      .p-inline-list{
+        bottom: 0;
+        position: absolute;
+      }
+    }
   }
 }
--- a/static/sass/_pattern_footer.scss
+++ b/static/sass/_pattern_footer.scss
@@ -3,13 +3,13 @@
     position: relative;
 
     .p-inline-list{
-      position: absolute;
       bottom: 0;
+      position: absolute;
     }
   }
 
   .p-list__item{
-    padding-top: 0;
     padding-bottom: 0;
+    padding-top: 0;
   }
 }
--- a/static/sass/_pattern_footer.scss
+++ b/static/sass/_pattern_footer.scss
@@ -1,44 +1,15 @@
 @mixin canonical-p-footer {
-  .p-footer {
-    @extend %vf-strip;
+  .row:first-child{
+    position: relative;
 
-    background-image: linear-gradient(
-        to bottom left,
-        rgba(119 41 83 / 100%) 0%,
-        rgba(119 41 83 / 100%) 49.9%,
-        rgba(119 41 83 / 0%) 50%,
-        rgba(119 41 83 / 0%) 100%
-      ),
-      linear-gradient(
-        to bottom left,
-        $color-accent-background 0%,
-        $color-accent-background 100%
-      );
-    background-position: top right;
-    background-repeat: no-repeat;
-    background-size: 25% 15%, 100%;
-    clip-path: polygon(100% 0%, 100% 100%, 0% 100%, 0% calc(0% + 4rem));
-    color: $color-mid-light;
-    padding-bottom: 2.5rem;
-    padding-top: 6rem;
-
-    &__link {
-      color: $color-mid-light;
-
-      &:hover,
-      &:visited {
-        color: $color-mid-light;
-      }
-    }
-
-    hr {
-      background: rgba(255 255 255 / 25%);
-      margin: 2rem 0;
+    .p-inline-list{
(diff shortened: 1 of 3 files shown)


lyubomir-popov:
The link colours are not readable, can we please use the ones in the design?
image

static/sass/_pattern_footer.scss
```diff
@@ -1,44 +1,15 @@
 @mixin canonical-p-footer {
-  .p-footer {
-    @extend %vf-strip;
+  .row:first-child{
```
anthonydillon, line 2: This is going to affect every `row` on the site. What is it for?

static/sass/_pattern_footer.scss
```diff
@@ -1,44 +1,15 @@
...
-      &:hover,
-      &:visited {
-        color: $color-mid-light;
-      }
-    }
-
-    hr {
-      background: rgba(255 255 255 / 25%);
-      margin: 2rem 0;
+    .p-inline-list{
```
anthonydillon, line 5: This will also affect every `p-inline-list` on the site. I would suggest if you are pushing something down you could create a new utility for this so its clear to others what is going on.

static/sass/_pattern_footer.scss
```diff
@@ -1,44 +1,15 @@
...
+      bottom: 0;
+      position: absolute;
     }
   }
-  @media (max-width: $breakpoint-x-small) {
-    .p-footer {
-      clip-path: polygon(130% 0%, 124% 134%, 0% 124%, 0% calc(0% + 4rem));
-    }
+
+  .p-list__item{
```
anthonydillon, line 11: You can get the same effect using `u-no-padding--top` and `u-no-padding--bottom`

anthonydillon (changes requested):
## QA
- [ ] The background of the footer should be `#262626`
- [ ] The positioning of the inline-list leads to strange layouts on small screens:
image

lyubomir-popov:
I'm looking into the color problem, and it is quite tricky. To pass  WCAG 2.0 rules about contrast against surrounding text, we are forced to use a colour that is difficult to read. Unless we decide it ok to fail the WCAG reqs to improve readability, I would suggest sticking to that  for now. Here are the values, I have asked Vanilla to make them available in the framework:
image


We will discuss this in the Accessibility guild and hopefully reach a conclusion whether we can afford to breach the outdated (but still in force) [WCAG requirements](https://www.w3.org/WAI/WCAG21/Techniques/general/G183), which is something most other sites do - including github and google chrome's settings pages for example.

For what is worth, following thenew APCA recommendations is also not ideal, because links become almost indistinguishable from white text, thereby failing the https://www.w3.org/WAI/WCAG21/Techniques/general/G183 technique.

static/sass/_pattern_footer.scss
mtruj013: Only for the footer, first 2 lines weren't meant to be deleted

static/sass/_pattern_footer.scss
mtruj013: Added the utility, lmk if this is what you meant!

static/sass/_pattern_footer.scss
mtruj013: But I'd have to add both to each list item, so I thought it'd be drier to just do it once. Lmk if you prefer the classes regardless and I'll change 🙂

mtruj013:
@anthonydillon @lyubomir-popov Thanks for the feedback! I've addressed your comments if you guys could take another look?

marinacastejon:
LGTM Thanks @mtruj013

anthonydillon (changes requested):
Thanks for the fixes @mtruj013 I have a few suggestions in the code to tidy it up and use Vanilla natively. Let me know if you have any questions.

anthonydillon (changes requested):
The background of the footer is now `#111111`

anthonydillon (approved):
LGTM :+1: thanks for the updates @mtruj013

templates/partial/_footer.html
```diff
@@ -1,42 +1,42 @@
 {% set current_path =  url_for(request.endpoint, **request.view_args)  %}
-<footer class="p-footer is-dark">
-    <div class="row">
-      <hr />
+<footer class="p-strip--dark p-footer">
+  <div class="row u-position">
+    <div class="col-3">
```
lyubomir-popov, line 4:
```suggestion
    <div class="col-medium-2 col-3">
```

templates/partial/_footer.html
```diff
@@ -1,42 +1,42 @@
...
-          </li>
-          <li class="p-inline-list__item">
-            <a href="https://twitter.com/Canonical" class="p-icon--twitter">Twitter</a>
-          </li>
-          <li class="p-inline-list__item">
-            <a href="https://www.youtube.com/user/celebrateubuntu" class="p-icon--youtube">YouTube</a>
-          </li>
-        </ul>
-      </div>
+    <div class="col-3">
```
lyubomir-popov, line 18:
```suggestion
    <div class="col-medium-2 col-3">
```

templates/partial/_footer.html
```diff
@@ -1,42 +1,42 @@
...
+      </ul>
     </div>
-    <div class="row">
-      <div class="col-12">
-        <p>
-          Ubuntu and Canonical are registered trademarks.<br />
-          All other trademarks are the property of their respective owners.
-        </p>
-      </div>
+    <div class="col-6">
```
lyubomir-popov, line 30:
```suggestion
    <div class="col-start-medium-3 col-6">
```

lyubomir-popov (changes requested):
Can you please review and accept my code suggestions? They improve the medium screen layout: 
image

lyubomir-popov:
Thanks, looking good!

PR #796, opened 2023-02-17
WD-2031 Add a no roles message to careers departments with no open roles

anthonydillon:
## Done
- Fixed the logic detecting and showing the no roles message
- Updated the message to match the Vanilla empty state styling
- Run prettier on the JS file

## QA

- Open the demo and go to /careers/all?filter=Administration
- See there is a message warning of no roles
- Click another department and see the message is removed and the roles are listed correctly.

## Screenshot
image

### DIFF
--- a/templates/partial/_careers-available-roles.html
+++ b/templates/partial/_careers-available-roles.html
@@ -9,7 +9,7 @@
         <li class="p-list__item is-ticked">Filtering by different department</li>
         <li class="p-list__item is-ticked">Trying a different spelling</li>
       </ul>
-      <a href="/careers/all" class="p-button--positive">View all roles</a>
+      <a href="/careers/all" class="p-button">View all roles</a>
     </div>
   </div>
 </div>
--- a/templates/partial/_careers-available-roles.html
+++ b/templates/partial/_careers-available-roles.html
@@ -1,7 +1,24 @@
 {% if vacancies %}
-<h2 class="p-heading--4 js-filter__no-results u-hide">
+<!-- <h2 class="p-heading--4 js-filter__no-results u-hide">
   Sorry, there are no open positions matching your criteria.
-</h2>
+</h2> -->
+<div class="p-strip js-filter__no-results u-hide">
+  <div class="row">
+    <div class="u-align--right col-3 col-medium-2 col-small-1">
+      <img src="https://assets.ubuntu.com/v1/0b82c071-World_icon.svg" alt="">
+    </div>
+    <div class="u-align--left col-5 col-medium-4 col-small-3">
+      <p class="p-heading--4 u-no-margin--bottom">No open positions matching your criteria</p>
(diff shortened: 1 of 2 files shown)


lyubomir-popov:
@anthonydillon that image is going to be contentious. Can we simply not have an image?

anthonydillon:
We can I just thought it made it nicer

anthonydillon:
Updated to this:
image

templates/partial/_careers-available-roles.html
```diff
@@ -1,7 +1,21 @@
 {% if vacancies %}
-<h2 class="p-heading--4 js-filter__no-results u-hide">
+<!-- <h2 class="p-heading--4 js-filter__no-results u-hide">
```
mtruj013, line 2: Left over comment?

mtruj013 (approved):
LGTM! Just a left over comment that we can delete I think

marinacastejon:
@anthonydillon a few comments:
- I agree with Lyubo's comment, much better without the image
- I think a better and more consistent copy would be: "There are no open roles matching your criteria."
- I see that this also appears in all empty states, could you please update the copy for all cases when there are no results?
Thanks

anthonydillon:
@marinacastejon believe I have addressed your points? Would you mind reviewing it again please?

marinacastejon:
LGTM, thanks @anthonydillon

lyubomir-popov:
@anthonydillon @marinacastejon do we need the filters when there's  nothing to filter? Why not simply do this instead: 
image

anthonydillon:
Applying filters could cause no roles so it would remove the ability to change that.

lyubomir-popov:
@anthonydillon ah I see. Then this? So no p-strip, and just one col offset, then inside - h2 instead of h4: 

image

anthonydillon:
@lyubomir-popov thanks, done.

templates/partial/_careers-available-roles.html
```diff
@@ -1,7 +1,18 @@
...
+  <div class="row">
+    <div class="u-align--left col-6 col-start-large-2">
+      <h2>There are no open roles matching your criteria</h2>
+      <p>Why not try widening your search? You can do this by:</p>
+      <ul class="p-list">
+        <li class="p-list__item is-ticked">Searching for alternative words or phrases</li>
+        <li class="p-list__item is-ticked">Filtering by different department</li>
+        <li class="p-list__item is-ticked">Trying a different spelling</li>
+      </ul>
+      <a href="/careers/all" class="p-button--positive">View all roles</a>
```
lyubomir-popov, line 12:
```suggestion
      <a href="/careers/all" class="p-button">View all roles</a>
```
we already have a primary button on this page

lyubomir-popov (approved):
All good, just one optional suggestion in the code.

PR #801, opened 2023-02-20
WD-2116 make search work with lower and upper case

pedoch:
## Done

- Fixed the search filter. Made it work with both lower-case and upper-case search values.

## QA

- Check out the demo on https://canonical-com-801.demos.haus/partners/find-a-partner and ensure the search filter works with lower-case and upper-case inputs, eg: `DELL`, `Dell`, and `dell` should all bring out search results for Dell Technologies.
- You can also test the filter for other partners as well. It should work for all.

## Issue / Card

- [WD-2116](https://warthogs.atlassian.net/browse/WD-2116)

[WD-2116]: https://warthogs.atlassian.net/browse/WD-2116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/static/js/find-a-partner.js
+++ b/static/js/find-a-partner.js
@@ -1,6 +1,6 @@
 (function () {
   const numberOfPartnersLabel = document.querySelector(
-    ".js-find-a-partner__number"
+    ".js-find-a-partner__number",
   );
   const searchBox = document.querySelector(".js-find-a-partner__search-input");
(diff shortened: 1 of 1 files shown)


tbille (approved):
works as expected :+1:

PR #810, opened 2023-03-02
Rebuild travel and lifestyle page

mtruj013:
## Done

- Applied redesign as per [design](https://www.figma.com/file/KQSCmTqOjY4gfKvJPyvzj2/Canonical.com-latest-only?node-id=7367%3A20953&t=Ghb8qeVFsNnSVkNO-0)

## QA

- View the site locally in your web browser at: https://canonical-com-810.demos.haus/careers/travel
- Check against redesign
- Note that the copy for this page is a combination of what were once 2 different pages: [travel](https://docs.google.com/document/d/1H2ILhAUckOWyah_6V69ImwLWOSRlXkaYnIhOHkUjclc/edit#heading=h.jyjo9jkeh22g) and [lifestyle](https://docs.google.com/document/d/15Ofb6vRUXbQ3evcWmrOu9ymuY37ayg4M_xZ2QXQyIts/edit#heading=h.jyjo9jkeh22g). So please check the final copy against how these were merged in the design

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-177


static/js/sprint-map.js
```diff
@@ -0,0 +1,81 @@
...
+    [{ lat: 55.67473557077814, lng: 12.602819433367 }, "Copenhagen"],
+    [{ lat: 50.09169724226367, lng: 14.37895031427894 }, "Prague"],
+    [{ lat: 50.142222694420674, lng: 8.614639914385569 }, "Frankfurt"],
+    [{ lat: 45.50800862995117, lng: -73.58280686860392 }, "Montreal"],
+    [{ lat: 40.798155988207235, lng: -73.85861239168297 }, "New York City"],
+    [{ lat: 25.019212561496353, lng: 121.46383373680085 }, "Taipei"],
+    [{ lat: 30.355265001335628, lng: -97.76687845245245 }, "Austin"],
+    [{ lat: 48.857074990278676, lng: 2.3395547447652056 }, "Paris"],
+    [{ lat: 29.767193226058257, lng: -95.3670538129764 }, "Houston"],
+    [{ lat: -31.043359293503908, lng: 25.51255683173218 }, "Spier"],
```
petesfrench, line 19: I couldn't find this one on google maps

static/js/sprint-map.js
```diff
@@ -0,0 +1,81 @@
...
+    [{ lat: -33.45421283631078, lng: -70.66577622308252 }, "Santiago"],
+    [{ lat: 39.748447785422016, lng: -105.0168199879092 }, "Denver"],
+    [{ lat: 30.05563678743736, lng: -89.82322663188988 }, "New Orleans"],
+    [{ lat: 52.24325667424824, lng: 21.05145033879068 }, "Warsaw"],
+    [{ lat: 42.446882419194324, lng: -71.22491732631751 }, "Lexington"],
+    [{ lat: 34.35676353830776, lng: -111.51505767860898 }, "Arizona"],
+    [{ lat: 25.763155578558653, lng: 54.94246340504953 }, "Dubai"],
+    [{ lat: 47.65006756574396, lng: -122.35867431899362 }, "Seattle"],
+    [{ lat: 37.42579354504801, lng: -78.17007396261319 }, "Virginia"],
+    [{ lat: 41.078424184612594, lng: -9.051132205419249 }, "Portugal"],
```
petesfrench, line 46: This pin looks a little off
image

petesfrench:
We should add a fallback for people who don't have JS enabled:
image

anthonydillon:
I am going on a limb and adding a design +1 as the page seems to match the Figma design. Checked on different viewports and a11y reviewed.

anthonydillon:
There are some a11y issues introduced by Google maps which do not seem fixable

templates/careers/travel.html
```diff
@@ -1,66 +1,112 @@
...
+      <div class="col-6">
+        <h3 class="p-heading--2">Travel is part of the deal</h3>
+      </div>
+      <div class="col-6">
+        <p>Travel isn't for everybody but if you are excited to see the world then Canonical is a great place to be. We expect everybody at Canonical to be willing to travel for up to two weeks at a time, four times a year, but in some roles the travel expectation is higher.</p>
+        <div class="u-hide--small u-hide--medium">
+          {{ 
+            image (
+            url="https://assets.ubuntu.com/v1/379b19b9-department-engineering-%402x.jpg",
+            alt="",
```
petesfrench, line 74: An alt text would be useful here

templates/careers/travel.html
```diff
@@ -1,66 +1,112 @@
...
+        <p>We also travel more than most companies, because our customers and our teams are global, and we need regular face-time with both.</p>  
+        <p class="p-heading--5">Third, we aspire to lead on the global stage.</p>   
+        <p>There's no other way to describe it than hard &ndash; hard work, hard challenges, hard competition. Most of us are avid students of greatness &ndash; we are interested in how the latest things work, how the best organizations work, how the smartest companies work, and we aim to hold our own in that crowd.</p>
+        <p>You will need to show that you can compete in deeply challenging intellectual fields. Your colleagues are aiming for the top, and they depend on you to set the same standard in your area of responsibility. That's tough. It's also incredibly satisfying, with continuous learning and challenge. You will have exceptional opportunities and exceptional colleagues. They will expect you to be exceptional too.</p>
+      </div>
+    </div>
+  </div>  
+</section>
+<script defer src="{{ versioned_static('js/sprint-map.js') }}"></script>
```
petesfrench, line 107: Do you think it is worth adding this to the build?

petesfrench (changes requested):

petesfrench (approved):

marinacastejon:
@mtruj013 LGTM, thanks

templates/careers/travel.html
mtruj013: I'm actually not sure. My first instinct was no because this isn't building something based on a dependency (like fuse js or latest news for example), but there's a very high likelihood that I'm totally wrong in that assessment 😅 @anthonydillon what would be best practice here?

mtruj013:
@anthonydillon @petesfrench @marinacastejon  Thanks all for reviewing! I think I've addressed most concerns but I have some questions.

**JS not enabled**
I've added what I think is a very hacky fix for when JS is not enabled and I'd like to get your feedback. I moved the styles to be added within the script instead so that if JS is not enabled the div doesn't take a defined width and height. That fixed the huge empty div situation. But now the question is, is this the behavior we want? Would we rather have a static image shown as opposed to showing nothing in place of the map? 

**Image alt text**
For the image with 4 locations I wasn't sure what to set as the alt text as I don't actually know what cities are in the image. I can see NYC, Seoul, and what looks like Budapest but I've no idea about the other one. Is "Sprint locations" good enough here? 

**Nested strips**
This point was raised in MM, but just to clarify, the nested p-strip convention comes from design and is on every single department page. If this is not best practice then we should clarify and document that.

anthonydillon:
Thanks for all the work on this @mtruj013. 

**JS not enabled**
I would suggest we could just render a list of locations in the markup for non-js in a `<nosrcipt>`.

**Image alt text**
It would be better to describe the image the best you can. You could even use something like "An image split in four showing four cities that we have visited, including NYC, Seoul, etc".

**Nested strips**
Sounds good, I will bring that up with Vanilla for guidance.

**a11y bugs**
Wave is complaining about the markers not having an alt. After a quick Google it looks like there is no quick fix.

tbille:
- Mobile view of the map has a fixed width which makes the scroll vertically :scream: 
image

- Same issue with JS disabled:
image

static/js/sprint-map.js
```diff
@@ -0,0 +1,32 @@
+function initMap() {
+  const map = new google.maps.Map(document.getElementById("sprint-map"), {
+    zoom: 3,
+    center: { lat: 38.44079856183539, lng: -32.13058355540764 },
+  });
+  
+  const mapDiv = document.getElementById("sprint-map");
+  mapDiv.style.height = "25rem";
+  // mapDiv.style.width = "69rem";
```
tbille, line 9: This line can be removed.

carkod (approved):
LGTM

PR #812, opened 2023-03-07
Respect the Greenhouse application answer order in select elements

anthonydillon:
## Done

- Reverse the rendering order on a single select dropdown in application answers.

## QA

- Go to /careers/3979744/application
- Check the order of the "How did you do in maths, physics or computer science at high school?" start with "Cannot recall" and end with "Top student"

## Screenshot
image


andesol (approved):
:+1:

PR #843, opened 2023-04-20
Get hiring leads bio

carkod:
## Done

- Get hiring leads bio info from [directory API](https://github.com/canonical/directory-api)
- Added test for this new function
- Refactored harvest call to reuse the one in `application/` module

### Decisions behind this PR
- I chose to use the directory API [employee filter query](https://github.com/canonical/directory-api/blob/main/schema.graphql#L70), because just by using it in the [graphql interface](https://directory.wpe.internal/graphql/) I noticed less time in the request than querying hundreds of employees, and then filtering by name. Currently it loops through a list of hiring leads, which is fine, because it's a very limited list of items in the static file.
- As a result, this also removed 3 loops in the code and one "if" condition. Because I removed the `hiring_leads_list` entirely and added all this new data to the `hiring_lead` object, thus removing a list(array) and creating and adding data to a flat object which contains all the greenhouse + directory API info. This makes it very easy to see all available info at one glance.
- Moreover, I've removed the reliance on matching `lead["name"] == application["hiring_lead"]["name"]`, because this is unreliable, instead I used the `employee_id`, which I assume it will exist 100% of the time, so no error handling was added in case of failure (the previous way of using names can be problematic, there could be 2 people with the same name, there could be people with several surnames, one person could have several emails, so email is also problematic). If at some point this fails, then it could mean we have a data consistency problem in the database? (missing unique ID for an employee)
- One drawback is that I was unable to loop through to check the existence of "video", so what I did instead is just match that single employee which has a video. If we are adding it somewhere in the future, we can revisit this by simply updating the `"hiring_lead"` object.

Alternatively, we could query the entire list of employees, in which case, there would be less changes, so I'd still add the list of employees to the `["hring_leads_list"]` object and everything else would be the same.

## QA

- Check https://canonical-com-843.demos.haus/careers/application/<token>, if you don't have a token, let me know.
- Check the new Bio section, it should look the same
- Check the "Withdraw" process works (I couldn't get to the very end, the toekn didn't seem to be valid, it must be an issue with the setting of my email in greenhouse)
- Test the video is showing by using Andreu's token.
- Please check any other page that you deem may be affected by this, this QA section is only what I see (I simply checked what uses `hiring_lead_list`), so I could be missing something that is breaking now.

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-1554?atlOrigin=eyJpIjoiZjNmZGU4OTQ5MWZhNDUxY2E3ODM2YmRhYzllZTZmYTQiLCJwIjoiaiJ9

## Screenshots


carkod:
This is just an initial PR, to decide on which type of query we want to do.

After that, we still have the indicated things: video and application withdrawal page

webapp/application.py
```diff
@@ -73,11 +76,44 @@
...
+def _get_employee_directory_data(employee_id: str):
+    # Provide a GraphQL query
+    transport = RequestsHTTPTransport(
+        url=directory_api_url,
+        headers={"Authorization": directory_api_token},
+        use_json=True,
+        verify=False,
+    )
+
+    client = Client(transport=transport, fetch_schema_from_transport=True)
```
tbille, line 96: Can you move the client to be initialised at the initialization of the application 

(line 81)

tbille:
Can you aslo update the konf to read the secrets, update the documentation to explain how to setup the graphql secrets?

webapp/application.py
carkod: Ok, so I am seeing a problem putting this outside. If no key is provided, the entire project will fail to run. `dotrun` will not work and the Github actions are going to require an API key. In the interview scheduler](https://github.com/canonical/interview-scheduler/blob/3534e41f84510201eafa4da69a45a7c1324808c2/scheduler/directory.py) we can see that it removes the depedency for poject initialization and only initializes it when the Directory API is required. In the case of careers, I don't see such a need, we only are using one endpoint, so there is no need for a class. So I think for now we can keep it inside of this function to remove the dependency?

tbille (approved):

carkod:
Trying to get the demo running... There may be something else going on.

andesol:
Maybe it works for you, but last time I tried to access the directory API (which is deployed in the WPE cluster) from web team's k8s I was getting some timeouts.

carkod:
> Maybe it works for you, but last time I tried to access the directory API (which is deployed in the WPE cluster) from web team's k8s I was getting timeouts.

It seems to work now

PR #858, opened 2023-05-11
WD-1154 Get hiring leads bio

carkod:
## Done

Reverts canonical/canonical.com#857

## QA

Use these links to check:
https://pastebin.canonical.com/p/5rd67JwfBF/

- First link is an actual application
- Second link is a dummy application for testing the only video we have, as we I don't have an application of that specific hiring lead.


tbille (approved):
LGTM

PR #882, opened 2023-06-06
WD-4421 Update application job name when transferring roles

anthonydillon:
## Done
1. Simplify the template logic by parsing a reliable job post name before passing it to the context.
2. Added logic to check for the applied job id matches the job post id. If it does its a post from the same job. If not, the role has been transferred and we need to render the job_name.
3. _Drive by_: converted the `elif` to a `else` as we want one to always be true. Which fixed a rendering issue in production.

## QA

- Open the demo and go to [these URLs](https://pastebin.canonical.com/p/3MnyxvkY3P/) to check they render as expected.
- Check that the first URL in the pastebin renders correctly after the assessment stage unlike production

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-4421

### DIFF
--- a/templates/careers/application/index.html
+++ b/templates/careers/application/index.html
@@ -1,6 +1,6 @@
 {% extends '/careers/base_job-details.html' %}
 
-{% block title %}{{ application["candidate"]["first_name"] }}, application for {{ application["jobs"][0]["name"] }}{% endblock %}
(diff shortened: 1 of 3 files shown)


tbille (approved):
looks good!
:ship: it!

PR #904, opened 2023-06-21
Add redirect from identity to diversity

anthonydillon:
## Done

Add redirect from identity to diversity

## QA

- Open the demo
- Go to /careers/diversity/identity
- See that it redirects to /careers/company-culture/diversity


edlerd (approved):
LGTM

PR #905, opened 2023-06-21
Redesign documentation landing and subpages

mtruj013:
## Done

- Applied redesign as per [mock-up](https://www.figma.com/file/KQSCmTqOjY4gfKvJPyvzj2/Canonical.com-latest-only?type=design&node-id=9999-32699&t=jovsgOJ31F3rH5s1-0)

## QA

- View the site locally in your web browser at: https://canonical-com-905.demos.haus/documentation
- Check each tab against design and see that it matches

## Issue / Card

Fixes

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -359,6 +359,21 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights, 
   }
 }
 
+.p-side-navigation__drawer.is-documentation{
+ ul li {
+  background: #f3f3f3;
+
+  a {
+    color: #3097ff;
+    padding-left: 0;
+
+    &:visited {
+      color: #AB83D3;
+    }
+  }
+ }
+}
+
 // XXX: Ant - 21/02/2023
 // Fix for breakage introduced in Vanilla v3.11.1
 // Reported: https://github.com/canonical/vanilla-framework/issues/4664
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -360,14 +360,14 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights, 
 
 .p-side-navigation__drawer.is-documentation{
  ul li {
-  background: #f3f3f3;
+  background: $color-paper;
 
   a {
-    color: #3097ff;
+    color: $color-link;
     padding-left: 0;
 
     &:visited {
-      color: #AB83D3;
+      color: $color-link-visited;
     }
   }
  }
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -1,4 +1,4 @@
-$color-tabs-active-bar: #000000;
+$color-tabs-active-bar: #000;
 @import "../../node_modules/vanilla-framework";
 
 // import cookie policy
@@ -374,9 +374,9 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights, 
 }
 
 .documentation-images {
+  height: 100px;
   object-fit: cover;
   width: 64px;
-  height: 100px;
 }
 
 // XXX: Ant - 21/02/2023
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -1,5 +1,4 @@
-$color-tabs-active-bar: #e95420;
-
+$color-tabs-active-bar: #000000;
 @import "../../node_modules/vanilla-framework";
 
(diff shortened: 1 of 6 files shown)


mtruj013:
@lyubomir-popov I'm having some trouble recreating this. The [Vanilla tabs](https://vanillaframework.io/docs/patterns/tabs) don't quite behave this way, the line and the tabs start at the same point. Is there another pattern I should be using here? 
Screenshot from 2023-06-21 16-20-15

ClementChaumel:
The dark bar on the left is a bit off
image

but I'm wondering @lyubomir-popov @juanruitina why is this not using the [table of content component](https://vanillaframework.io/docs/patterns/table-of-contents)? or at least the [Side navigation](https://vanillaframework.io/docs/patterns/navigation#side-navigation)

static/sass/styles.scss
```diff
@@ -359,6 +358,27 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights,
...
+  a {
+    color: #3097ff;
+    padding-left: 0;
+
+    &:visited {
+      color: #AB83D3;
+    }
+  }
+ }
+}
```
ClementChaumel, line 374: We should probably use a different pattern altogether but if we're going forward with this let's use vanilla colours!

static/sass/styles.scss
```diff
@@ -1,5 +1,4 @@
-$color-tabs-active-bar: #e95420;
-
+$color-tabs-active-bar: #000;
```
ClementChaumel, line 1: I don't think this is used?

static/sass/styles.scss
```diff
@@ -359,6 +358,27 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights,
...
+ ul li {
+  background: #f3f3f3;
+
+  a {
+    color: #3097ff;
+    padding-left: 0;
+
+    &:visited {
+      color: #AB83D3;
+    }
```
ClementChaumel, line 371:
```suggestion
  background: $color-paper;

  a {
    color: $color-link;
    padding-left: 0;

    &:visited {
      color: $color-link-visited;
    }
```

ClementChaumel (approved):

ClementChaumel:
> @lyubomir-popov I'm having some trouble recreating this. The [Vanilla tabs](https://vanillaframework.io/docs/patterns/tabs) don't quite behave this way, the line and the tabs start at the same point. Is there another pattern I should be using here? Screenshot from 2023-06-21 16-20-15

If we're going forward with this hr, we should push it 1px to the top 
image

(But I think the actual fix is to just use the vanilla tabs

lyubomir-popov:
> but I'm wondering @lyubomir-popov @juanruitina why is this not using the [table of content component](https://vanillaframework.io/docs/patterns/table-of-contents)? or at least the [Side navigation](https://vanillaframework.io/docs/patterns/navigation#side-navigation)

I had used a list of anchors in the design, because this isn't a sidenav - but let's just use the default sidenav, seems like less effort from where things curently are

ClementChaumel:
> nav - but let's just use the default sidenav, seems like less effort from where things curently are

In function, it is a table of content. Why are we not using the existing pattern, are we deprecating the vanilla component?

static/sass/styles.scss
mtruj013: It is, it's the bottom border color for active tabs here: 
Screenshot from 2023-06-22 11-46-38

with the old color:
Screenshot from 2023-06-22 11-46-47

lyubomir-popov:
> > nav - but let's just use the default sidenav, seems like less effort from where things curently are
> 
> In function, it is a table of content. Why are we not using the existing pattern, are we deprecating the vanilla component?

You're right. Let's just use that. I needs improving but I can do that as a vanilla pr and it will then update here. I'll put it on my list.

juanruitina:
@lyubomir-popov We already have this open: https://github.com/canonical/vanilla-framework/issues/4777 Let's work together on it.

lyubomir-popov:
@mtruj013 should we have breadcrumbs here? cc @juanruitina

juanruitina:
I don't think we need breadcrumbs here, they would be too redundant with the tabs.

ClementChaumel:
I'm switching the target to the refresh branch. Undo if it's not correct

lyubomir-popov:
Should we close this one @mtruj013? I think we agreed to continue work on the other one

PR #910, opened 2023-06-24
Wd 4674 visual update partners pages footer

lyubomir-popov:
## Done

Revert to previous design as per discussion with @juanruitina 

## QA

- Check that the partners footer is in a 2 6-col layout, check links go to the correct places

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-4674

## Screenshots
image

### DIFF
--- a/templates/partners/iot-device.html
+++ b/templates/partners/iot-device.html
@@ -126,6 +126,6 @@
   </div>
 </section>
 
-{% include 'partial/_partners-footer.html' %}
+{% include '/partners/partial/_footer.html' %}
 
 {% endblock %}
--- a/templates/partners/partial/_footer.html
+++ b/templates/partners/partial/_footer.html
@@ -1,34 +1,30 @@
-  <section class="p-section">
-    <div class="row">
-      <hr class="p-rule">
-      <div class="col-6 col-medium-3">
-        <h3 class="p-heading--2">
-          Find a partner
-        </h3>
-      </div>
-      <div class="col-6 col-medium-3">
-        <p>Looking for a supplier with Ubuntu expertise? Select from the complete list of our partners.</p>
-        <hr class="p-rule">
-        <p>
-          <a href="/partners/find-a-partner">Find a partner&nbsp;&rsaquo;</a>
-        </p>
-      </div>
+<section class="p-section">
+  <div class="row">
+    <hr class="p-rule">
+    <div class="col-6 col-medium-3">
+      <h3 class="p-heading--2">Find a partner</h3>
     </div>
-  </section>
-  <section class="p-strip is-deep u-no-padding--top">
-    <div class="row">
+    <div class="col-6 col-medium-3">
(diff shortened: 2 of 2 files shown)


templates/partial/_partners-footer.html
```diff
@@ -1,9 +1,32 @@
-<section class="p-strip--white is-deep">
-  <div class="row">
+<section class="p-section">
+  <hr class="p-rule is-fixed-width">
+  <div class="row p-section">
```
mtruj013, line 3: In the [Vanilla documentation](https://vanillaframework.io/docs/patterns/section) it says that sections should not be nested inside of each other, am I understanding that correctly?

lyubomir-popov: Yes, I don't know what is the best course for this. We need this kind of spacing, how we obtain it is a good question. We could either update the vanilla docs, or create a new component or modifier class. cc @bartaz.

The basic idea is this: depending on context, some things might get lost or not have the necessary "gravitas". This is when we use the bigger of the two spacings we have. (The smaller being `p-block`).

mtruj013: IMO if there are cases where we're supposed to ignore the documentation, then the documentation should be updated to clearly reflect that

bartaz: If we are putting a section inside section to achieve double bottom margin than it's clearly something wrong.

All it is I believe is a "deep" spacing. So this should not be a double section but a deep strip.
And as far as I remember we discussed that when you need a deep spacing then it's needed both on top and bottom of given part of content (because it needs same "emphasis" from both sides), that's why the deep spacing is only available in deep strips.

So seems like the page content above the footer should be wrapped in a deep strip if we want a deep spacing between content and footer. But I guess we will again end up in a point there this strip will only need deep spacing on bottom and not on top (which is what we wanted to avoid with new components)… So we seem to be running in circles a bit again.

Allowing nesting sections in each other is a red flag for me, and it should be the last resort. In this case I believe this can be solved by using deep spacing (so a deep strip instead). I guess the remaining question for @lyubomir-popov is, is this still true that with deep spacing we always need it both on top or bottom, or are there cases (like this one likely?) when we only need deep spacing on the bottom. If so, we need to add deep section I guess?

anthonydillon: If the reason for this is to create a visible disconnect between the content and the footer. Could we simply add margin-top to footers across all sites?

juanruitina:
For some reason footer hasn't change in the [IoT page](https://canonical-com-910.demos.haus/partners/iot-device), is it hardcoded perhaps?

lyubomir-popov:
My guess is it isn't rebased on a branch that has that change yet, which is normal

lyubomir-popov:
This is ready for another look @mtruj013 @juanruitina.

I've used deep strip with no padding top to avoid contradicting the vanilla documentation. And I've replaced the hardcoded partners footer (well spotted @juanruitina) on the iot page with the correct partial.

ClementChaumel (approved):
Super duper 🤙

PR #920, opened 2023-06-27
fix marketo payload fields

laszlokajtar:
https://sentry.is.canonical.com/canonical/ubuntu-com/issues/29389


anthonydillon (approved):
LGTM but we still need the field `grecaptcharesponse` on the form in Marketo.

laszlokajtar:
thanks @anthonydillon, added that to the marketo form already

PR #931, opened 2023-07-05
WD-4982 - Make the navigation tag orange

anthonydillon:
## Done

Update the accent color on the site which is used for the navigation tag by Vanilla.

## QA

- Open the demo and see the logo is orange and not aubergine anymore

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-4982

## Screenshots

image

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -513,7 +513,7 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights, 
   background-size: contain;
 }
 
-.p-department-cards__container {  
+.p-department-cards__container {
   aspect-ratio: auto;
   background-color: #fff;
 
@@ -546,7 +546,7 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights, 
 
     .p-department-cards__link {
       color: $color-link;
-      
+
       &:hover {
         text-decoration: underline;
       }
@@ -558,7 +558,7 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights, 
       top: 0.4375rem;
       width: 2rem;
     }
-    
+
     .p-department-cards__image-container {
       align-items: center;
       display: flex;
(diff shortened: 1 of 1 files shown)


static/sass/styles.scss
```diff
@@ -700,3 +701,7 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights,
 #latest-articles article:first-child hr {
   display: none;
 }
+
+.p-navigation__tagged-logo .p-navigation__logo-tag {
+  background-color: #E95420;
```
bartaz, line 706: Does it need both variable change and override here?

static/sass/styles.scss
```diff
@@ -1,4 +1,5 @@
 $color-tabs-active-bar: #e95420;
+$color-accent: #e95420;
```
lorumic, line 2: I think this is already defined in Vanilla, so maybe it is superfluous to re-define it here?

static/sass/styles.scss
lorumic, line 706: Would it be good to reference the variable here instead of the hard-coded value?
```suggestion
  background-color: $color-accent;
```

bartaz:
I didn't know that canonical with alpha version of 4.0 is already live.

Vanilla 4.0 uses `$color-brand` for logo, not accent color, so it's the brand variable that would need changing, but I guess on this version of Canonical website we are using aubergine.

Anyway, now that we have this bug live it's fine to override, but it will be fixed in Vanilla soon as well.

static/sass/styles.scss
bartaz: I agree, this seems unnecessary.
```suggestion
```

bartaz:
BTW, this is already fixed in Vanilla 4.0 branch, will be published with next release: https://github.com/canonical/vanilla-framework/pull/4821

carkod:
@lyubomir-popov you may want to take a look at this

anthonydillon:
Good catches @bartaz @lorumic I have updated this now just to temp fix until we have Vanilla 4.0+

PR #935, opened 2023-07-05
Hide fast tracks

anthonydillon:
## Done
Hide fast tracks


carkod (approved):

carkod:
We could also hide everything else except Dev Rel, but from a design perspective would not look good.

PR #954, opened 2023-07-12
Update /press-centre headings

mtruj013:
## Done

- Updated headings as per ticket

## QA

- Don't think it's needed here, since there should be no changes visually

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-5029


ClementChaumel (approved):
Super duper 👏️

PR #966, opened 2023-07-17
Disable cache

mtruj013:
## Done

- Set max cache time to an hour as per stakeholder request

## QA

- See that the site is accessible? 

## Issue / Card

Fixes #

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -683,7 +683,7 @@ def cache_headers(response):
     import ipdb
     ipdb.set_trace()
     if flask.request.path.startswith(disable_cache_on):
-        response.cache_control.max_age = "600"
+        response.cache_control.max_age = "3600"
 
     return response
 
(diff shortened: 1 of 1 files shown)


webapp/app.py
```diff
@@ -665,6 +665,28 @@ def allow_src(tag, name, value):
...
+        "/documentation",
+        "/projects"
+    )
+
+    if flask.request.path.startswith(disable_cache_on):
+        response.cache_control.max_age = "3600"
+
+    return response
+
+
```
carkod, line 689: I believe python linting is complaining about additional space.

carkod (approved):
LGTM but I don't think this disables the cache? Just reduces it to 60 seconds like you said. Disabling would be setting to `no-cache` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

PR #968, opened 2023-07-18
Clean up custom styles

mtruj013:
## Done

- Removed any custom styles that are not needed or are currently not being used
- Drive by: 
  - Update strip usage on /partners/executive-summit 

## Note

There are still some styles to remove here, specifically those used in templates that no longer belong on the site. To not bloat this PR too much I will remove these in a separate PR along with custom styles they use 

## QA

- View the site locally in your web browser at: https://canonical-com-968.demos.haus/
- Click through the site and make sure nothing is broken

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-4915


templates/partners/executive-summit.html
```diff
@@ -6,8 +6,10 @@
 
 {% block meta_description %}Connect with partners from around the globe to learn about open source, from data centre to the cloud, to maximize your company sales.{% endblock %}
 
+{% set is_paper = True %}
```
carkod, line 9: Another thing we could do here is instead of having `is_paper` in the [base_index.html](https://github.com/canonical/canonical.com/blob/eee7f73d656e65942f384138b8f74246d7f7221b/templates/base_index.html#L83) we should maybe change it to set a class. I am suggesting this because you guys have added the `is_grapphite` body class, so maybe setting a class instead of true or false is a better option moving on. https://github.com/canonical/ubuntu.com/blob/6b288d3e210f483fb382e8338ff694bed8a90054/templates/templates/base.html#L70

carkod (approved):
Well done! Thanks. I don't see any classes that conflict with my currently ongoing PRs, so I gave +1

templates/partners/executive-summit.html
carkod: Or even on step further would be to set all the pages with paper white background (as long as design is consistent with future designs) by default and adding classes to override

carkod:
Something else we should do in another PR probably, is to do this run but with templates. Search templates that are not being used (no `includes`).

mtruj013:
> Something else we should do in another PR probably, is to do this run but with templates. Search templates that are not being used (no `includes`).

Yup! I mentioned this in the notes section of the pr description, I have a separate ticket for that 🙂

templates/partners/executive-summit.html
mtruj013: Thanks @carkod, great shout about this. I went with the first option since we still have a few pages that are not supposed to be using the paper background

lyubomir-popov:
@mtruj013  Sorry to comment on a closed pr, but is this used? 
image
Seems to still be in main after the cleanup. If we are using it, we have probably missed a page that needs to be converted to the new styles that do not use this kind of surus.

mtruj013:
@lyubomir-popov please read the [pr description](https://github.com/canonical/canonical.com/pull/968#issue-1810147803), I'm covering the rest in a separate PR

carkod:
Yeah let's do it in another PR. Reviewing 36 files is not a lot of fun.

> @mtruj013 Sorry to comment on a closed pr, but is this used? image Seems to still be in main after the cleanup. If we are using it, we have probably missed a page that needs to be converted to the new styles that do not use this kind of surus.

Looking at [this](https://github.com/search?q=repo%3Acanonical%2Fcanonical.com%20p-strip--suru&type=code) yes we are still those in quite a few pages.

mtruj013:
> seems like there's still a lot of things that can be cleaned up

> Looking at [this](https://github.com/search?q=repo%3Acanonical%2Fcanonical.com%20p-strip--suru&type=code) yes we are still using those in quite a few pages. Althought a lot of them look like old pages that are unused.

Guys I am begging you to read the pr description, I took the time to write it so we could avoid this unnecessary chasing. 

"**There are still some styles to remove here, specifically those used in templates that no longer belong on the site. To not bloat this PR too much I will remove these in a separate PR along with custom styles they use**"

Meaning: I am aware that there are styles left to remove. I am aware that there are templates left to remove. I am addressing those changes in a separate PR to make it easier to review.

PR #985, opened 2023-07-25
Fix typos in interview-done card

CarsonHoffman:
## Done

- Multiple typos in `_interview-card-done.html` fixed.

I didn't think that filing an issue would be appropriate for such a small change, but I can do so if that would be helpful. 🙂


anthonydillon (approved):
Thanks for fixing these. LGTM 👍

PR #1090, opened 2023-10-25
Build MicroCloud landing

mtruj013:
## Done

- Added microcloud landing page as per [design](https://www.figma.com/file/hzQYJM9Tui4wpmyZC0k0WS/MicroCloud-Rebranding?node-id=0%3A1&mode=dev) and [doc](https://docs.google.com/document/d/17iBbFCINLYjPBOYWpgoq9SreyYNkR0Sb5jgx5iTo6MA/edit)

NOTE: Any CTA's that are supposed to route to either the modal or stand-alone form for this bubble will not work as they have not been built yet.

## QA

- View the site locally in your web browser at: http://0.0.0.0:8002/microcloud
- Check against doc and design

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-6692

### DIFF
--- a/templates/microcloud/index.html
+++ b/templates/microcloud/index.html
@@ -1,0 +1,435 @@
+{% extends 'base_index.html' %}
+
+{% block title %}MicroCloud &ndash; low-touch open source cloud solution{% endblock %}
+
+{% block meta_description %}MicroClouds offer a lightweight and automated solution for edge clouds and small scale private clouds. Best of open source cloud infrastructure components integrated seamlessly.{% endblock %}
+
+{% block meta_copydoc %}https://docs.google.com/document/d/17iBbFCINLYjPBOYWpgoq9SreyYNkR0Sb5jgx5iTo6MA/edit{% endblock meta_copydoc %}
+
+{% block body_class %}is-paper{% endblock body_class %}
+
+{% block content %}
+
+<section class="p-strip--whitesuru is-shallow">
+  <div class="row">
+    <div class="col-start-large-4 col-9">
+      <div class="p-section--shallow">
+        <h1>The power of cloud<br/>anywhere you need it</h1>
+      </div>
+      <div class="p-section--shallow">
+        <p>Straightforward and lightweight, but featureful.<br/>MicroClouds serve compute near consumers, wherever they may be.</p>
+      </div>
+      <a class="p-button--positive">Get in touch</a>
+      <a href="https://ubuntu.com/engage/guide-to-micro-clouds">Read our guide to MicroCloud</a>
+    </div>
+  </div>
+</section>
+
+<section class="p-section">
+  <div class="row--25-75">
+    <hr class="p-rule"/>
+    <div class="col">
+      <h2>What is a MicroCloud?</h2>
+    </div>
+    <div class="col">
+      <div class="row">
+        <div class="col-6">
+          <div class="u-embedded-media">
+            <iframe class="u-embedded-media__element" title="What is a MicroCloud?" width="560" height="315" src="https://www.youtube.com/embed/FnzUcGg60Qk?si=CdadVAcSZGLHLi6S" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+</section>
+
+<section class="p-section">
+  <div class="u-fixed-width p-strip">
+    <hr class="p-rule"/>
+    <h2>Deploy a fully functional cloud in minutes:</h2>
+    <div class="row-cli__commands">$ <span class="row-cli__commands-list p-heading--2 typer" id="main"
+      data-words="snap install microcloud lxd microceph microovn,microcloud init"
+      data-delay="50" data-deletedelay="5000" data-colors="#111111"
+      style="color: rgb(17, 17, 17);">snap install mic</span></div>
+    </div>  
+  </div>
+  <div class="row">
+    <div class="col-start-large-4 col-9 col-medium-6">
+      <div class="p-section--shallow">
+        <div class="row">
+          <div class="col-3 col-medium-2 p-section--shallow">
+            {{ 
+              image (
+              url="https://assets.ubuntu.com/v1/c459b7cd-microcloud-visual-1.svg",
+              alt="",
+              width="100",
+              height="100",
+              hi_def=True,
+              loading="lazy"
+              ) | safe
+            }}
+          </div>
+          <div class="col-3 col-medium-2 p-section--shallow">
+            {{ 
+              image (
+              url="https://assets.ubuntu.com/v1/273d81f3-microcloud-visual-2.svg",
+              alt="",
+              width="83",
+              height="100",
+              hi_def=True,
+              loading="lazy"
+              ) | safe
+            }}
+          </div>
+          <div class="col-3 col-medium-2 p-section--shallow">
+            {{ 
+              image (
+              url="https://assets.ubuntu.com/v1/a8264def-microcloud-visual-3.svg",
+              alt="",
+              width="100",
+              height="100",
+              hi_def=True,
+              loading="lazy"
+              ) | safe
+            }}
+          </div>
+          <hr class="p-rule"/>
+          <div class="col-3 col-medium-2 p-section--shallow">
+            <p class="p-heading--5">Automated deployment</p>
+            <p>MicroClouds are built with simplicity and automation in mind. With MicroClouds, both deployment and on-site operations are automated with low to no on-site maintenance, regardless of whether you're running it at the back of the office or at a hard-to-reach remote edge site.</p>
+          </div>
+          <div class="col-3 col-medium-2 p-section--shallow">
+            <p class="p-heading--5">Secure with hassle-free updates</p>
+            <p>While smaller in scale, your MicroClouds will still likely process big amounts of data. They may also run sensitive workloads. Security and prompt vulnerability updates are the key priorities to ensure your data is kept safe. MicroCloud components are packaged using snaps, providing an extra layer of isolation while enabling worry-free updates.</p>
+          </div>
+          <div class="col-3 col-medium-2 p-section--shallow">
+            <p class="p-heading--5">Cost and resource efficient </p>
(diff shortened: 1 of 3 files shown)


juanruitina:
Two very minor requests:

* Please make "Automated deployment", "Secure with hassle-free updates", "Cost and resource efficient" H3s.
* Change "Small-scale private cloud" to "Scalable private cloud" by Miona's request.

This looks good UX-wise otherwise, feel free to UX+1 once you fix that. Thank you!

lyubomir-popov:
Hi @mtruj013, looks good overall, a few small issues below:

Spacing should be 24 at the top, 64 at the bottom. This is the case for most heroes, see for example ubuntu.com/data:
image

This is not styled as our code element, please see the vanilla example
image

this p-strip class is not needed:
image

normal p-section here, not shallow:
image

p-rule above the heading here (sorry was missing in the design):
image

this p-section class is not needed:
image

can we please wrap this heading in a p-section--shallow:
image


can we please have this section retain the internal layout on medium screens:
image

mtruj013:
@lyubomir-popov ready for another look, let me know if I missed something

lyubomir-popov:
hi @mtruj013, much better, a few small things still remain though:

- the code element styling still doesn't match the vanilla example or the design - maybe there are old  styles from the previous implementation?
- the space under it has now become far too small - see comparison here (it used to be too big, but now it has almost disappeared. the correct value is one p-section wrapper adding 4rem of spacing at the bottom of that element): 
image

-the button here needs to be outside this p-section--shallow, the idea is to insert a shallow space in between heading and button:
image

- thhis `<hr>` breaks the grid - elements with no classes are treated as 12 columns, while the parent is 9 columns. best to move it outside the row:
- 
image

- another broken grid, probably for a similar reason:
image

-space after the deep strip, the div hihglighted in blue is not needed IMO 
image

mtruj013:
@lyubomir-popov could you take another look? I tried my best to get the animation closer to the design but if it's still not up to par could you be specific about what needs to change? Color, font, font-weight, spacing, etc. 

> -space after the deep strip, the div hihglighted in blue is not needed IMO

I've removed the `p-section` class from the parent section, please let me know if that's what you meant. 

Also, please note that in general a lot of spacing isn't correctly reflected in the design, or at least not as clearly as it has been in the past. A deep strip should show spacing on both top and bottom in dev mode, the design only shows spacing on the bottom which denotes a `p-section`:
image

Compare against the microk8s mock-ups for example where it is clear that this is meant to use a p-strip regular: 
image

I'll flag this feedback in whatever retro is most relevant so it doesn't get lost in the shuffle, but just wanted to make a note for the future.

lyubomir-popov:
Hi @mtruj013, 

The animation looks spot on!

Some clarificaitons on  your other questions:

> I've removed the `p-section` class from the parent section, please let me know if that's what you meant.

Added some fixes so you can see, there were a few unnecessary levels of nesting.

> Also, please note that in general a lot of spacing isn't correctly reflected in the design, or at least not as clearly as it has been in the past. A deep strip should show spacing on both top and bottom in dev mode, the design only shows spacing on the bottom which denotes a `p-section`:

The microk8s example is not exactly an apples to apples comparison - the section here contains a small caps heading. This is why in figma it is marked up like so:

outer container - deep section:

image

then inside, shallow p-section with small caps inside:

image

followed by a shallow p-section with the h2 inside:

image

and finally, a button after all this.

I'm open to suggestions on how to make it clearer, let's talk in the next retro.

Please see my commit, it addresses all my concerns. I hope I haven't broken anything in your code. 
I've added Design +1.

templates/microcloud/index.html
```diff
@@ -0,0 +1,444 @@
...
+    <div class="col">
+      <h2>What is the size of a MicroCloud?</h2>
+    </div>
+    <div class="col">
+      <div class="row">
+        <div class="col-3">
+          <p>MicroClouds are highly available by default, so a minimum of three systems are needed.</p>
+          <p>They can easily extend in site size and scale from a small three-node to larger 50-node clusters.</p>
+        </div>
+        <div class="col-3 p-image-wrapper">
```
petesfrench, line 136:
```suggestion
        <div class="col-3 p-image-wrapper u-align--center">
```
This is horizontally centered in the design

templates/microcloud/index.html
```diff
@@ -0,0 +1,444 @@
...
+      <h2>Key features</h2>
+    </div>
+    <div class="col">
+      <ul class="p-list--divided">
+        <li class="p-list__item is-ticked">Highly available cluster deploys in minutes.</li>
+        <li class="p-list__item is-ticked">Automatic security updates and streamlined upgrades.</li>
+        <li class="p-list__item is-ticked">Runs virtual machines, or system containers for higher density.</li>
+        <li class="p-list__item is-ticked">Remote authentication, RBAC, projects and multi-user setup.</li>
+        <li class="p-list__item is-ticked">Easily replicated at scale.</li>
+        <li class="p-list__item is-ticked">Fully open-source, commercial support available with <a href="https://ubuntu.com/pro">Ubuntu Pro</a></li>
```
petesfrench, line 220:
```suggestion
        <li class="p-list__item is-ticked">Fully open-source, commercial support available with <a href="https://ubuntu.com/pro">Ubuntu Pro</a>.</li>
```

templates/microcloud/index.html
```diff
@@ -0,0 +1,444 @@
...
+      <p>
+        <a href="https://ubuntu.com/engage/guide-to-micro-clouds">Learn more about MicroClouds for the edge&nbsp;&rsaquo;</a>
+      </p>
+    </div>
+  </div>
+  <div class="row--50-50 p-section--shallow">
+    <hr class="p-rule"/>
+    <div class="col">
+      <div class="p-section--shallow">
+        <h3 class="p-heading--2">Scalable private cloud</h3>
```
petesfrench, line 267:
```suggestion
        <h3 class="p-heading--2">Scalable private clouds</h3>
```

templates/microcloud/index.html
```diff
@@ -0,0 +1,444 @@
...
+        </div>
+      </div>
+    </div>
+  </div>
+</section>
+
+<section class="p-section">
+  <div class="u-fixed-width">
+    <hr class="p-rule"/>
+    <h2>Deploy a fully functional cloud in minutes:</h2>
```
petesfrench, line 52: I wonder if we should hide this from screen readers. Everything gets read correctly, but without visual queues is there much value in it? What do you think @juanruitina?

templates/microcloud/index.html
```diff
@@ -0,0 +1,444 @@
...
+        </div>
+      </div>
+    </div>
+  </div>
+</section>
+
+<section class="p-section">
+  <div class="u-fixed-width p-section">
+    <hr class="p-rule"/>
+    <h2>Best-of-breed<br/>open-source components</h2>
```
petesfrench, line 156: I think we need to look at new ways to break lines, as this causing some problems with screen readers (atleast with Orca), wherein it sees the two lines as separate and will only read one at a time. I have added this as an item in the a11y guild

templates/microcloud/index.html
```diff
@@ -0,0 +1,444 @@
...
+      </ul>
+    </div>
+  </div>
+</section>
+
+<section class="p-section">
+  <div class="u-fixed-width">
+    <hr class="p-rule"/>
+    <h2 class="p-text--small-caps">MicroCloud use cases</h2>
+    <div class="">
```
petesfrench, line 230:
```suggestion

```

templates/microcloud/index.html
carkod: We can add the `off-screen` or `aria-hidden`

templates/microcloud/index.html
```diff
@@ -0,0 +1,443 @@
...
+    <div class="row">
+      <div class="col-start-large-4 col-9">
+        <div class="p-section--shallow">
+          <h1>The power of cloud<br/>anywhere you need it</h1>
+        </div>
+        <div class="p-section--shallow">
+          <p>Straightforward and lightweight, but featureful.<br/>MicroClouds serve compute near consumers, wherever they may be.</p>
+        </div>
+        <hr>
+        <a class="p-button--positive">Get in touch</a>
```
carkod, line 24: This modal/link is not working

templates/microcloud/index.html
carkod: Very cool effect. If we are not hiding it, can we make tweaks to the responsive smaller scree sizes? I think the text needs to be smaller in smaller screen sizes, the cursor is not aligning correctly. But not a block in my opinion or we can just hide like Pete says

Screenshot from 2023-11-01 09-58-10

templates/microcloud/index.html
```diff
@@ -0,0 +1,443 @@
...
+      <p class="p-text--small-caps">All the benefits of a cloud, but localised</p>
+    </div>
+    <div class="p-section--shallow">
+      <h2>MicroClouds are low-touch, efficient and reliable.<br/> Do you have a use case in mind? Get in touch with our team.</h2>
+    </div>
+        <a class="p-button--positive">Talk to an expert</a>
+    </div>
+</section>
+
+<script src="/static/js/typer.js" defer></script>
```
carkod, line 442: Should we use `versioned_static`? or do you issues with it?

templates/microcloud/index.html
```diff
@@ -0,0 +1,443 @@
...
+          </div>
+          <div class="col-3 col-medium-2 p-section--shallow">
+            <h3 class="p-heading--5">Secure with hassle-free updates</h3>
+            <p>While smaller in scale, your MicroClouds will still likely process big amounts of data. They may also run sensitive workloads. Security and prompt vulnerability updates are the key priorities to ensure your data is kept safe. MicroCloud components are packaged using snaps, providing an extra layer of isolation while enabling worry-free updates.</p>
+          </div>
+          <div class="col-3 col-medium-2 p-section--shallow">
+            <h3 class="p-heading--5">Cost and resource efficient </h3>
+            <p>In addition to standard Ubuntu Server or Desktop, MicroClouds can also be run on Ubuntu Core -  a lightweight OS optimised for the edge. Combined with Ubuntu Core, users get a truly lightweight cloud, giving you resource efficiency without compromising on performance, while keeping costs at bay.</p>
```
carkod, line 115: This looks a bit weird. There is an additional space before "a" and you may want to replace it with `&dash;`?

carkod (approved):
A few suggestions, LGTM

templates/microcloud/index.html
mtruj013: I added an aria-hidden to is as it doesn't read both commands, only the one currently being typed. I'll make a follow-up for design on smaller screens though!

templates/microcloud/index.html
petesfrench: Unfortunately both of these don;t do anything for the screen reader, so we will have to explore alternatives

PR #1097, opened 2023-10-30
Add resources page and slider

mtruj013:
## Done

- Built page as per [doc](https://docs.google.com/document/d/1ffTxn2ymKrWRPUjjx6cN0-OA6rlIuX0YaVbFQHcpsTM/edit) and [design](https://www.figma.com/file/KQSCmTqOjY4gfKvJPyvzj2/Canonical.com-latest-only?node-id=12964%3A15505&mode=dev)

## QA

- View the site locally in your web browser at: https://canonical-com-1097.demos.haus/microcloud/resources
- Check against copy and design

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-6695

### DIFF
--- a/templates/microcloud/resources.html
+++ b/templates/microcloud/resources.html
@@ -24,7 +24,6 @@
     <hr class="p-rule"/>
     <div class="col">
       <h2 class="p-text--small-caps" id="videos-and-webinars">Videos and webinars</h2>
-      <a class="p-text--small-caps" href="https://ubuntu.com/engage?tag=microcloud" aria-label="Explore all videos and webinars">Explore all</a>
     </div>
     <div class="col">
       <div class="row js-slider--items">
@@ -84,13 +83,17 @@
         </div>
       </div>
       <hr class="p-rule"/>
-      <button class="js-slider--prev" title="Previous resources"> 
-        <i class="p-icon--chevron-down"></i>
-      </button>
-      <span id="js-results-count"></span>
-      <button class="js-slider--next" title="Next resources"> 
-        <i class="p-icon--chevron-down"></i>
-      </button>
+      <div class="row">
+        <div class="col-9">
+          <button class="js-slider--prev" title="Previous resources"> 
+            <i class="p-icon--chevron-down"></i>
+          </button>
+          <span id="js-results-count"></span>
+          <button class="js-slider--next" title="Next resources"> 
+            <i class="p-icon--chevron-down"></i>
+          </button>
+          <a class="p-text--small-caps explore-link-align" href="https://ubuntu.com/engage?tag=microcloud" aria-label="Explore all videos and webinars">Explore all</a>
+        </div>
     </div>
   </div>
 </section>
--- a/templates/microcloud/resources.html
+++ b/templates/microcloud/resources.html
@@ -103,10 +103,9 @@
     <hr class="p-rule"/>
     <div class="col">
       <h2 class="p-text--small-caps">Case studies,<br aria-hidden="true"/>whitepapers and reports</h2>
-      <a class="p-text--small-caps" href="https://ubuntu.com/engage?tag=microcloud" aria-label="Explore all case studies, whitepapers and reports">Explore all</a>
     </div>
     <div class="col">
-      <div class="row">
+      <div class="row p-section">
         <div class="col-3">
           <div class="p-section--shallow">
             <h3 class="p-heading--5">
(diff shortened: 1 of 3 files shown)


juanruitina:
Couple of comments:

- Let's make the pagination's text refer to the number of pages, so: 1 of 2, 2 of 2.
- On small/medium, there's a weird jump further down when clicking on the pagination. In these cases when items are stacked, let's make the page scroll to the top (so, to "Videos and webinars") instead.
- It's great that we make the items/pages/tabs always the same height, makes interaction very smooth. However, I noticed it doesn't adjust well on small/medium. Given that they are stacked on those viewports, can we just limit that behaviour to desktop only?

mtruj013:
@lyubomir-popov could you take a look if/when you get a chance?

juanruitina:
I noticed some differences with the mockup:

- Can you please move the "Explore all" link right-aligned on the pagination bar? (This might be a recent tweak to the design)
- Also, we should add pagination to the blog section, given we have four items now.

Minor tweak: please add "blog posts" after "Explore all" in the link label for screens only so we make that link distinct to screen readers.

mtruj013:
@juanruitina Sorry, I don't understand what you mean by this:

> Can you please move the "Explore all" link right-aligned on the pagination bar as per the mockup? (This might be a recent tweak to the design)

Current version:
Screenshot from 2023-11-03 12-54-47

Mock-up:
Screenshot from 2023-11-03 12-55-05

The link alignment already matches the mock-up. If by pagination bar you mean this, then I'm still confused as there is nothing right aligned there either (current mock-up as of 1pm German time): 
image

mtruj013:
@juanruitina I added aria-labels to all the "explore more" links, but I'm going to leave the pagination for the blog posts as an enhancement. I wasn't told that the task was to make it so that if there are more than 3 items per section it should have pagination so I didn't build it with that in mind. I would need to refactor it in order for the click not to trigger pagination in both sections, but since this version already matches the design I was given I will leave that for a follow-up PR. Please let me know if you're ok with that!

juanruitina:
Just noticed the Figma file is the wrong one :/ It was [this](https://www.figma.com/file/hzQYJM9Tui4wpmyZC0k0WS/23.10-C.com---MicroCloud-Rebranding?type=design&node-id=4%3A2&mode=design&t=6g9ZBvlLdgUbuZAK-1). This is what it should look like:

Screenshot 2023-11-03 at 15 38 16

I see what you mean about the blog posts, please keep track of it so we can tackle it later. Thank you for taking care of the aria-labels, hadn't noticed!

mtruj013:
@juanruitina could you take a look again? @lyubomir-popov could you also take a look?

mtruj013:
@lyubomir-popov

lyubomir-popov:
Hi @mtruj013 I am sorry about the confusion with design links. This is [final](https://www.figma.com/file/hzQYJM9Tui4wpmyZC0k0WS/23.10-C.com---MicroCloud-Rebranding?type=design&node-id=4%3A2&mode=design&t=qr0hjeAjpYG639xk-1) Looks great, just 3 comments:

- can we ensure there's no jump as you paginate? One way, set a maximum character count on the description and ellpsise when it is exceeded, or set the entire section's height based on the largest element within it. 
- this shouldn't be small caps, regulat link works: 
image
- can we please put all explore all links under a hr on the right, same as the first one - otherwise it is inconsistent:
image

mtruj013:
@lyubomir-popov updated! Let me know if there's anything else

lyubomir-popov:
Hi @mtruj013 much better, still a few niggles though:
- small caps here:
image

these should be in paragraphs - otherwise they don't get the padding-top that comes from paragraphs, and they sit too closely to the line above:
image

- I would hide this line on small screens, as it looks inconsistent with the preceding items:
image

- can we please keep the slider functionality on medium screens - something like this:
image

mtruj013:
@lyubomir-popov update, could you take another look?

templates/microcloud/resources.html
```diff
@@ -0,0 +1,188 @@
+{% extends 'base_index.html' %}
+
+{% block meta_description %}A collection of useful and educational material and information about MicroClouds and its components.{% endblock %}
```
petesfrench, line 3: We should add a meta title

petesfrench (approved):

juanruitina:
LGTM!

PR #1188, opened 2024-02-26
auto delete interviews on withdrawal (WD-7721)

nathanclairmonte:
## Done

- Added Google Calendar API class
  - Class taken from the [interview scheduler repo](https://github.com/canonical/interview-scheduler/blob/main/scheduler/google_calendar.py)
  - Added a `delete_event_from_interview_calendar()` function which as you might guess deletes an interview from the interview calendar (the interview calendar is the shared calendar on which all interviews are scheduled)
  - Also added `sendUpdates="all"` to the delete request itself. This ensures a notification is sent to all attendees of the meeting telling them that it was canceled. The old parameter for this was `sendNotifications` (a boolean), but that has since been deprecated.
- Updated the email sent to the hiring lead when a candidate withdraws. Before, it explicitly told them to look for any pending interviews and delete them, but now this isn't necessary anymore since it will happen automatically.
- Created a new template for the email that will be sent to each interviewer letting them know that their interview was canceled because the candidate withdrew their application.
- Added debug mode functionality to the main `/templates/application/withdrawal.html` template (which is displayed once the candidate withdrawal is confirmed) to display the emails that would have been sent to the interviewers if not in debug mode.
- Implemented the actual interview auto-deletion flow:
  - When the candidate clicks on the confirmation link they received in their email to confirm their withdrawal, the `application_withdrawal()` (line 465 in `/webapp/application.py`) function is called.
  - Inside this function, we pull all the scheduled interviews for the current application from the Harvest API and filter for the ones with a status of `scheduled` (the other options for this status are `complete` and `awaiting_feedback`, both of which indicate that the interview has already happened and thus does not need to be deleted).
  - We loop through this list of scheduled interviews, grab the interviewer info and their timezone, format the interview date/time to match the interviewers timezone (so that the email they receive shows the correct time for them), delete the interview, and then create the email to alert the interviewer of the cancelation.
  - Each cancelation interview is saved in an array so that we can display them all in debug mode.
  - **UPDATE:** We are now also pulling the interviews with a status of `awaiting_feedback` so that we can alert those interviewers that a scorecard submission is no longer necessary. These interview events are not deleted however, as they will have already passed.

### Prep
- Add `SERVICE_ACCOUNT_EMAIL` and `SERVICE_ACCOUNT_PRIVATE_KEY` to `.env.local` (for Google Calendar API)
- Create a test candidate on GH for the "Dev - test 1" job. Make sure to input your email address (or one you have access to) since it will be needed to set the candidate's availabilities
- Move the test candidate to the "Early Stage Interviews" stage.
- Add yourself as a default interviewer for this candidate.
- Input candidate availabilities via email specifically. I.e. click "Request Availability", send the email, fill out the form received in email.
- Wait for main interview-scheduler cronjob to run (it runs every 4 hours) or run the test-scheduler cronjob, then an interview should be scheduled.
  - **NB:** I also have a local script that can schedule these interviews the same way the interview-scheduler does, so if you aren't getting one scheduled when it runs (this can happen sometimes if it just doesn't find one to schedule) or if it's just taking too long, let me know and I can manually schedule a test interview for you.
- Make sure the interview actually shows up in GH, otherwise it won't show up in the scheduled interviews returned from the Harvest API.
- Once there is a scheduled interview in GH, you can move on to the steps below to run the site and test the deletion process.

### Testing the auto-deletion flow
- Run the site with `dotrun`
- View the site locally in your web browser at: http://localhost:8002/
- Navigate to the test candidate's candidate page (you can get the candidate page URL from the `https://harvest.greenhouse.io/v1/applications/{{application_id}}` endpoint on the Harvest API)
- Click the withdraw button, enter the email you put as the candidate's email, select any reason and click submit
- On the subsequent page, you should see a debug message along with the withdrawal confirmation email that would have been sent to the candidate. Click that link to confirm the withdrawal.
- This should reject the candidate on GH, delete any upcoming interviews, and then render the `/templates/application/withdrawal.html` template, which should have the email(s) that would've been sent to the interviewer(s).

(btw, let me know if this is too long! I tried to be as like thorough as possible with the QA steps)

## Issue / Card

[WD-7721](https://warthogs.atlassian.net/browse/WD-7721)

## Screenshots

[WD-7721]: https://warthogs.atlassian.net/browse/WD-7721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/careers/application/_withdrawal_notification-email.html
+++ b/templates/careers/application/_withdrawal_notification-email.html
@@ -2,7 +2,7 @@
   <p>Dear {{ hiring_lead_name }},</p>
   <p>This is an automated email to inform you that {{ applicant_name }} has withdrawn their application for the {{ position }} position.</p>
   <p>The candidate was in {{ current_stage["name"] }} stage.</p>
-  <p>If this is an interview stage, please make sure you check the activity feed of the candidate in Greenhouse and delete any pending interviews directly from Google Calendar.</p>
+  <!-- <p>If this is an interview stage, please make sure you check the activity feed of the candidate in Greenhouse and delete any pending interviews directly from Google Calendar.</p> -->
   <p><a href="{{application_url}}">Click here to access the candidate's application.</a></p>
   <p>Regards,<br>Talent Science Engineering</p>
 </div>
--- a/templates/careers/application/_withdrawal_notification-email.html
+++ b/templates/careers/application/_withdrawal_notification-email.html
@@ -2,7 +2,6 @@
   <p>Dear {{ hiring_lead_name }},</p>
   <p>This is an automated email to inform you that {{ applicant_name }} has withdrawn their application for the {{ position }} position.</p>
   <p>The candidate was in {{ current_stage["name"] }} stage.</p>
-  <!-- <p>If this is an interview stage, please make sure you check the activity feed of the candidate in Greenhouse and delete any pending interviews directly from Google Calendar.</p> -->
   <p><a href="{{application_url}}">Click here to access the candidate's application.</a></p>
   <p>Regards,<br>Talent Science Engineering</p>
 </div>
--- a/webapp/application.py
+++ b/webapp/application.py
@@ -77,7 +77,6 @@ application = flask.Blueprint(
 
 session = talisker.requests.get_session()
 harvest = Harvest(session=session, api_key=os.environ.get("HARVEST_API_KEY"))
-calendar = CalendarAPI()
 cipher = Cipher(os.environ.get("APPLICATION_CRYPTO_SECRET_KEY"))
 base_url = "https://harvest.greenhouse.io/v1"
 
@@ -496,6 +495,7 @@ def application_withdrawal(token):
     ]
     all_sent_emails = []
 
+    calendar = CalendarAPI()
     for scheduled_interview in scheduled_interviews:
         # get interviewer information
         interviewer = scheduled_interview["interviewers"][0]
--- a/webapp/application.py
+++ b/webapp/application.py
@@ -487,15 +487,6 @@ def application_withdrawal(token):
         f"application_id={payload['application_id']}"
     )
 
-    # call the Harvest API to reject the application
-    response = harvest.reject_application(
-        application["id"],
-        application["hiring_lead"]["id"],
-        withdrawal_reason_id,
-        withdrawal_message,
-    )
-    response.raise_for_status()
-
     # get all scheduled interviews
     scheduled_interviews = harvest.get_interviews_scheduled(application["id"])
     scheduled_interviews = [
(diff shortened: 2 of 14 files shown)


webapp/application.py
```diff
@@ -484,6 +486,90 @@ def application_withdrawal(token):
...
+        interviewer_timezone = calendar.get_timezone(interviewer["email"])
+
+        # convert interview time to interviewer's timezone
+        date_time_str = interview["start"]["date_time"]
+        date_time_obj = datetime.fromisoformat(
+            date_time_str.replace("Z", "+00:00")
+        )
+        date_time_obj = date_time_obj.astimezone(
+            pytz.timezone(interviewer_timezone)
+        )
```
muhammadbassiony, line 511: instead of `date_time_str` and `date_time_str` lets name it to `interview_start_time_str` so we can stick to verbose variables names as much as possible

webapp/google_calendar.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+    def get_single_event(self, calendar_id, event_id):
+        return (
+            self.service.events()
+            .get(calendarId=calendar_id, eventId=event_id)
+            .execute()
+        )
+
+    def delete_event(self, calendar_id, event_id):
```
muhammadbassiony, line 66: Let's delete unneeded functions for now

muhammadbassiony (approved):
Amazing work 🚀 especially with the tests 🔥

carkod:
Just a note that we usually follow on Sites, would be nice not to have all those commits that are not useful like "Fix lint", "Cleanup print statements", they don't provide any value to the Git history. Thanks

webapp/google_calendar.py
```diff
@@ -0,0 +1,60 @@
+from google.oauth2 import service_account
+from googleapiclient.discovery import build
+from googleapiclient.errors import HttpError
+
+import os
```
carkod, line 5: Ideally should be at the top. https://peps.python.org/pep-0008/#imports.

You can use a plugin like isort, which is available in the vscode extensions to automatically do this for you.

templates/careers/application/_withdrawal_notification-email.html
```diff
@@ -2,7 +2,7 @@
   <p>Dear {{ hiring_lead_name }},</p>
   <p>This is an automated email to inform you that {{ applicant_name }} has withdrawn their application for the {{ position }} position.</p>
   <p>The candidate was in {{ current_stage["name"] }} stage.</p>
-  <p>If this is an interview stage, please make sure you check the activity feed of the candidate in Greenhouse and delete any pending interviews directly from Google Calendar.</p>
+  <!-- <p>If this is an interview stage, please make sure you check the activity feed of the candidate in Greenhouse and delete any pending interviews directly from Google Calendar.</p> -->
```
carkod, line 5: If we no longer need this, would be best to remove it completely. We have Git history always to take back content that we've removed

nathanclairmonte:
@carkod Thank you for the info and suggestions! I wasn't aware of this git command, will utilise it in future.

I've also addressed the import organization in `google_calendar.py` and the commented-out line in `_withdrawal_notification-email.html`.

Thanks again 🙏

PR #1189, opened 2024-02-27
Add the Enterprise Sales Representative video to the candidate page

anthonydillon:
## Done

Add the Enterprise Sales Representative video to the candidate page
Could not use the hiring JSON as this is on the hiring lead level not the individual role
Updated the title text

## QA

- Check the code looks right as the position is not open yet.
- Check some candidate pages to check they continue to work as expected

### DIFF
--- a/webapp/application.py
+++ b/webapp/application.py
@@ -251,9 +251,9 @@ def _get_application(application_id):
                 print(error)
 
             if job_id == "2680006":  # Enterprise Sales Representative
-                application["hiring_lead"]["video_src"] = (
-                    "https://www.youtube.com/embed/UvDSXgPbpt8"
-                )
+                application["hiring_lead"][
+                    "video_src"
+                ] = "https://www.youtube.com/embed/UvDSXgPbpt8"
             elif (
                 # Currently only user with video
                 # as we don't have a source to pull this video from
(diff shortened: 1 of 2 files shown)


akbarkz:
@anthonydillon Is there an open position currently that we could use to test this change? I couldn't find one.

anthonydillon:
Nope, as mentioned in the QA steps this position is not open yet. I'll try and assign a test candidate to the position but not sure I can if it's not open

webapp/application.py
```diff
@@ -250,7 +250,11 @@ def _get_application(application_id):
             except HTTPError as error:
                 print(error)
 
-            if (
+            if job_id == "2680006":  # Enterprise Sales Representative
```
akbarkz, line 253: What's the chance of this ID changing in the future? Are we sure it will always have this ID?

akbarkz (approved):

carkod:
@anthonydillon can you squash your "Fix lint" commits? They are not very useful in Git history.

webapp/application.py
anthonydillon: It's the id of the role we want to display. There is no system of role-based videos only HL at this point so its hacky but might just be a one off.

PR #1224, opened 2024-03-20
Add djLint formatter/linter for Jinja templates

akbarkz:
## Done

- Added djLint fotmatter/linter support
- Defined a configuration file `.djlintrc` for it
- Added a yarn script `lint-jinja` that checks linting in Jinja templates that were changed
- Added GH CI job for running `lint-jinja` script

## QA

- Set up djLint in your codebase as described in Discourse topic here: https://discourse.canonical.com/t/configuring-editors-for-formatting-and-linting/188
- Open any Jinja template and format it
- Save the formatted file
- Run `yarn lint-jinja` script in the terminal and check that it has checked one changed file and that the check has passed
- Change another Jinja template and save it without formatting it
- Run `yarn lint-jinja` and check that the file is reported as having a linting error
- Open the file and check where djLint is reporting errors by hovering on places that are highlighted red
- Format the file and fix remaining linting errors and save
- Run `yarn lint-jinja` again and check that the file has passed the linting check

## Issue / Card

There is no issue

### DIFF
--- a/.djlintrc
+++ b/.djlintrc
@@ -1,0 +1,18 @@
+{
+  "blank_line_after_tag": "load,extends",
+  "blank_line_before_tag": "load,extends,block, macro",
+  "close_void_tags": true,
+  "ignore": "J018,H006,H021,H023,H029,T002,T003,T028",
+  "include": "H017",
+  "max_blank_lines": 1,
+  "max_line_length": "120",
+  "profile": "jinja",
+  "format_css": true,
+  "format_js": true,
+  "css": {
+    "indent_size": 2
+  },
+  "js": {
+    "indent_size": 2
+  }
+}
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -98,7 +98,7 @@ jobs:
           CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
         run: |
           echo "The following files have changed: $CHANGED_FILES"
-          yarn lint-jinja
+          djlint $CHANGED_FILES --lint
 
   test-python:
     runs-on: ubuntu-22.04
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -86,8 +86,19 @@ jobs:
           python3 -m pip install --upgrade pip
           sudo pip3 install djlint
 
+      - name: Get changed HTML files in the templates folder
+        id: changed-files
+        uses: tj-actions/changed-files@v43
+        with:
+          files: templates/**
+
       - name: Lint jinja
-        run: yarn lint-jinja
+        if: steps.changed-files.outputs.any_changed == 'true'
+        env:
+          CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          echo "The following files have changed: $CHANGED_FILES"
+          yarn lint-jinja
 
   test-python:
     runs-on: ubuntu-22.04
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -72,6 +72,23 @@ jobs:
       - name: Lint python
         run: yarn lint-python
 
+  lint-jinja:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install node dependencies
+        run: yarn install --immutable
+
+      - name: Install python dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          sudo pip3 install djlint
+
+      - name: Lint jinja
+        run: yarn lint-jinja
+
   test-python:
     runs-on: ubuntu-22.04
 
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "format-prettier": "prettier -w 'static/js/*.{js,jsx,ts,tsx}' 'static/sass/*.scss'",
     "lint-python": "flake8 --extend-ignore=E203 webapp tests && black --check --line-length 79 webapp tests",
     "lint-scss": "stylelint static/**/*.scss",
+    "lint-jinja": "djlint $(git diff --name-only HEAD | xargs) --lint || true",
     "serve": "./entrypoint 0.0.0.0:${PORT}",
     "start": "yarn run build && concurrently --raw 'yarn run watch' 'yarn run serve'",
     "test": "yarn run lint-scss && yarn run lint-python && yarn run test-python",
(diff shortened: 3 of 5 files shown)


.djlintrc
```diff
@@ -0,0 +1,18 @@
+{
+  "blank_line_after_tag": "load,extends",
+  "blank_line_before_tag": "load,extends,block, macro",
+  "close_void_tags": true,
+  "ignore": "J018,H006,H021,H023,H029,T002,T003,T028",
```
bartaz, line 5: Out of curiosity, what are these codes?

.github/workflows/pr.yaml
```diff
@@ -72,6 +72,23 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install node dependencies
+        run: yarn install --immutable
+
+      - name: Install python dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          sudo pip3 install djlint
```
bartaz, line 87: If djlint is in `requirements.txt` as a dependency, does it need to be installed separately here?

.github/workflows/pr.yaml
```diff
@@ -72,6 +72,23 @@ jobs:
...
+
+      - name: Lint jinja
+        run: yarn lint-jinja
```
bartaz, line 90: It seems that `lint-jinja` script only checks modified files (`git diff --name-only HEAD`), so would that even work on CI when everything is commit, there are no local changes?

.djlintrc
akbarkz: Sure, they are listed here: https://www.djlint.com/docs/linter/#rules
Basically they are codes for certain linting rules

.github/workflows/pr.yaml
akbarkz: you're right, it doesn't need to be in `requirements.txt`, I will remove

bartaz: But if it's removed from `requirements.txt` it will not be installed in development environment, right? So devs will have to install it manually?

I believe it should stay in `requirements.txt`, but maybe in CI we can install dependencies from requirements.txt instead of manually installing djlint?

I'm not sure how it works with other Python based linting, like black.

bartaz: Hmm… ok it seems that we are installing black on CI directly, so I guess it's fine to do the same with djlint

https://github.com/canonical/canonical.com/blob/13a3265e6f140c8811eb4053a60006a7a9a3367b/.github/workflows/pr.yaml#L70

bartaz: It looks like black and flake8 are part of dotrun image, so that's why they are installed by default in dev environment.

Until djlint is added to dotrun, I think it needs to be in requirements.txt of a project, so it is installed in dev environment:

https://github.com/canonical/dotrun-image/blob/63768e54385e465c2a2e29155613d3d7e313eb06/src/setup.py#L13-L20

.github/workflows/pr.yaml
akbarkz: that's a good point, it's useful to have this script locally before you have committed changes. But after you have committed it's not going to work. My idea was to run this check in CI only for files from the PR, but not for the whole codebase. Because in the team we decided to apply linting step by step, only to files that are being changed.

bartaz: Found something like this: https://github.com/marketplace/actions/changed-files

.github/workflows/pr.yaml
akbarkz: ok leaving it there then

.github/workflows/pr.yaml
akbarkz: thanks, I also find this https://github.com/codeocelot/git-diff-lint/blob/master/git-diff-lint, so currently playing around and checking if that works for us

akbarkz: @bartaz I ended up applying your finding and made it work nicely now. I will check all changed files inside the `templates` folder and lint them. You can check that it works by looking at the logs of the latest CI GH action below called `lint-jinja`. I changed one HTML file inside `templates` folder and you can see that it's being linted.

package.json
```diff
@@ -15,6 +15,7 @@
     "format-prettier": "prettier -w 'static/js/*.{js,jsx,ts,tsx}' 'static/sass/*.scss'",
     "lint-python": "flake8 --extend-ignore=E203 webapp tests && black --check --line-length 79 webapp tests",
     "lint-scss": "stylelint static/**/*.scss",
+    "lint-jinja": "djlint $CHANGED_FILES --lint || true",
```
bartaz, line 18: This will not run locally right? Because `$CHANGED_FILES` is only set on CI?

Maybe it doesn't have to run locally, if you don't want or don't need it to, but in such case, maybe it's not needed in package.json at all?
If it depends on `$CHANGED_FILES`, can only be run on CI, just run it directly in CI job. Because having it in package.json suggests it's something that can be run locally in dev environment, which currently it can't.

Also, why is the `|| true` needed?

akbarkz: yes, makes total sense @bartaz. I removed it from package.json.
And I don't know what `|| true` is for, it was in Andreu's similar yarn script, so I just copied it :D

akbarkz: Locally it's not needed because djlint will automatically format the code on save and will highlight any errors in files.

.djlintrc
```diff
@@ -0,0 +1,18 @@
...
+  "include": "H017",
```
bartaz, line 6: It's probably out of the scope of this PR, but maybe something we discuss at dev catchup to decide on the rules we want to enable/disable.

Judging on our experience with prettier it's easier and better to just follow the defaults and adjust our own internal rules to follow. So if the default is not to close void tags, maybe we can accept it and change our HTML guidelines, so that we don't have to make exceptions for something that is accepted as common standard.

akbarkz: Okay, shall I bring up this topic to the dev catchup meeting again, but this time go through each rule that was added/ignored?
I'm just afraid that we might end up never having it cause the discussion will keep going infinitely.
What about we merge it the way it is now, and if later someone is uncomfortable about a certain rule, they can bring it up to discussion themselves?

bartaz: I don't mean to block this on having the dev catchup discussion.

Let's merge this, start using it. But before we make it as a cross team tool with some shared config would be good to discuss is we need any exceptions from defaults.

bartaz (approved):
LGTM.
Thanks for doing this. This will be a good start. Let's try it out and see if there are any adjustments needed later.

PR #1236, opened 2024-04-08
Update cookie policy to v3.6.3

samhotep:
## Done

- Updated the coookie policy to v3.6.3

## QA

- View the demo in an incognito web browser at: https://canonical-com-1236.demos.haus/
- **Before** accepting cookies, 
  - Open developer tools and in the console type in `dataLayer`
  - There should be a `{"consent", "default"}` object in the dataLayer
- **After** accepting cookies, 
  - There should be a `{"consent", "update"}` object in the dataLayer

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "cypress:open": "cypress open --config-file tests/cypress/cypress.json --config baseUrl=http://0.0.0.0:${PORT:-8002}/"
   },
   "dependencies": {
-    "@canonical/cookie-policy": "3.6.0",
+    "@canonical/cookie-policy": "3.6.1",
     "@canonical/latest-news": "1.5.0",
     "@canonical/global-nav": "3.6.2",
     "@testing-library/cypress": "9.0.0",
--- a/package.json
+++ b/package.json
@@ -24,9 +24,9 @@
     "cypress:open": "cypress open --config-file tests/cypress/cypress.json --config baseUrl=http://0.0.0.0:${PORT:-8002}/"
   },
   "dependencies": {
-    "@canonical/cookie-policy": "3.6.1",
-    "@canonical/latest-news": "1.5.0",
+    "@canonical/cookie-policy": "^3.6.3",
     "@canonical/global-nav": "3.6.2",
+    "@canonical/latest-news": "1.5.0",
     "@testing-library/cypress": "9.0.0",
     "autoprefixer": "10.4.17",
     "concurrently": "7.6.0",
--- a/templates/base_index.html
+++ b/templates/base_index.html
@@ -1,149 +1,213 @@
-<!doctype html>
+<!DOCTYPE html>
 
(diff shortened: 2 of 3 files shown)


laszlokajtar:
hey @samhotep, i did a quick test and the results are much better, however, GTM still seems to load earlier than the default consent:

image
image

lorumic:
> * **After** accepting cookies,
>   
>   * There should be a `{"consent", "update"}` object in the dataLayer

Is the object with "consent", "default" in it supposed to go away after accepting the cookies? It currently stays there, alongside the one with "consent", "update".

samhotep:
> > * **After** accepting cookies,
> >   
> >   * There should be a `{"consent", "update"}` object in the dataLayer
> 
> Is the object with "consent", "default" in it supposed to go away after accepting the cookies? It currently stays there, alongside the one with "consent", "update".

Yeah it stays there by default

laszlokajtar:
@samhotep another issue is that the update seems to somehow work in reverse way,
in another test, i only allowed essential cookies, which resulted in this update push:

`{"ad_storage":"granted","ad_user_data":"granted","ad_personalization":"granted","analytics_storage":"granted","functionality_storage":"denied","personalization_storage":"denied","security_storage":"granted"}`

samhotep:
> @samhotep another issue is that the update seems to somehow work in reverse way, in another test, i only allowed essential cookies, which resulted in this update push:
> 
> `{"ad_storage":"granted","ad_user_data":"granted","ad_personalization":"granted","analytics_storage":"granted","functionality_storage":"denied","personalization_storage":"denied","security_storage":"granted"}`

This happens because the default object has all values denied, so any selection, even essential will cause some update since some values will be updated to `true`

laszlokajtar:
> > @samhotep another issue is that the update seems to somehow work in reverse way, in another test, i only allowed essential cookies, which resulted in this update push:
> > `{"ad_storage":"granted","ad_user_data":"granted","ad_personalization":"granted","analytics_storage":"granted","functionality_storage":"denied","personalization_storage":"denied","security_storage":"granted"}`
> 
> This happens because the default object has all values denied, so any selection, even essential will cause some update since some values will be updated to `true`

my bad, wrong selection choice, sorry please discard!

akbarkz (approved):

laszlokajtar:
@samhotep on this demo, GTM still loads first for me

laszlokajtar:
> @samhotep on this demo, GTM still loads first for me

is it possible that 

cpNs.cookiePolicy();

doesn't return before GTM is loaded?

samhotep:
> > @samhotep on this demo, GTM still loads first for me
> 
> is it possible that
> 
> cpNs.cookiePolicy();
> 
> doesn't push the required values into the dataLayer before GTM is loaded?

You're right, I've made the changes in the cookie policy, thanks!

PR #1100, opened 2023-10-31
Fix application job post id

carkod:
## Done

Fix [Jira ticket](https://warthogs.atlassian.net/browse/WD-6775) /careers/5172749

Swapped the Greenhouse job-boards API endpoint with Harvest API job post endpoint. The GH job boards API doesn't return job posts that have been flagged `inactive` or `not live`

Most of the files changed are due to the different data structure that this API returns, but the data seems to match.

## QA

- Go to https://canonical-com-1100.demos.haus/careers/5172749 should work now
- Button to apply should not show on the page (the harvest endpoint will return 401 upon submission of application)

Regression check:

- Go to e.g. https://canonical-com-1100.demos.haus/careers/4417916
- Click "Apply now" and fill out the application
- Go to Greenhouse and check that the application submitted successfully and correctly (check for missing fields, content submitted in the wrong fields etc, it's a bit hard for me to map the data without seeing the final result in greenhouse).

- Run through the following [QA steps](https://discourse.canonical.com/t/qa-steps/152)
- [List additional steps to QA the new features or prove the bug has been resolved]

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-6775

PR #1260, opened 2024-05-28
Fix missing requisition ID

carkod:
## Done

Append back the original content with Requisition ID, as requested.

#### Rationale
Originally this was changed to [show active/inactive roles](https://github.com/canonical/canonical.com/pull/1100), but looks like there is no way to get the Requisition ID with only the Harvest Job board API, so had to put back the old greenhouse API.

Also, since it looks like role description looks the same, so I directly assigned the old `job["content"]`.

## QA

- Go to https://canonical-com-1260.demos.haus/careers/4754075 and check that "Requisition ID" shows up now, at the end of the page (before the application button).
- Anything else you may think could be breaking with this change.

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-8203

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -260,11 +260,17 @@ def careers_rss():
     "/careers/<regex('[0-9]+'):job_id>/<job_title>", methods=["GET", "POST"]
 )
 def job_details(job_id, job_title):
+    """
+    job_title is not used, but is included in the route to avoid
+    breaking existing links
+    """
     context = {"bleach": bleach}
 
     try:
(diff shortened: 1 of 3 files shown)


webapp/app.py
```diff
@@ -288,8 +294,6 @@ def job_details(job_id, job_title):
                 "text": f"{response.reason}. Please try again!",
             }
 
-        return flask.render_template("/careers/job-detail.html", **context)
```
carkod, line 291: Redundant code

mtruj013 (approved):
LGTM!

PR #1418, opened 2024-11-07
specify timeouts on all http requests

mariuskava:
## Done

Specify timeouts on all http requests. 15s for get requests, 30s for post requests. It is a good practice to always set timeouts on network operations. In case a request hangs it might cause a connections/ram leak. It might also cause stale cache pages if cache validation request happens to hang.

## QA

- Check everything works as before. (careers pages, blog pages, partners pages)


mtruj013 (approved):
LGTM!

PR #1474, opened 2024-12-20
[WD-17951] use different footer for docs pages

akbarkz:
## Done

Moved a footer that was implemented for multipass.run's Docs pages some time ago into canonical.com and used it for all Docs pages on canonical.com.

## QA

- Open demo instance
- Open Docs pages, for example at /multipass/docs or /mircostack/docs and check that they use their own footer
- Open other pages (not Docs) and check that they still use an existing footer (similar to production)

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-17951

## Screenshots

Normal footer:
Screenshot From 2024-12-20 11-18-52

Docs footer:
Screenshot From 2024-12-20 11-19-16

### DIFF
--- a/templates/docs/shared/_footer.html
+++ b/templates/docs/shared/_footer.html
@@ -1,0 +1,40 @@
+<div class="p-strip is-dark l-docs__footer">
+  <div class="row">
+    <footer class="l-docs__subgrid">
+      <div class="l-docs__sidebar">
+        <p style="padding-left: 1.5rem">
+          Copyright &copy; {{ now("%Y") }}
+          <br />
+          CC-BY-SA 3.0, Canonical Ltd.
+        </p>
+      </div>
+      <div class="l-docs__main">
+        <nav aria-label="Footer">
+          <div class="has-cookie">
+            <ul class="p-inline-list--middot">
+              <li class="p-inline-list__item">
+                <a aria-label="External link to the Ubuntu legal information page"
+                   href="https://www.ubuntu.com/legal">Legal information</a>
+              </li>
+              <li class="p-inline-list__item">
+                <a class="js-revoke-cookie-manager" href="">Manage your tracker settings</a>
+              </li>
+              <li class="p-inline-list__item">
+                <a href="https://github.com/canonical/multipass/issues/new/choose">Report a bug on Multipass itself</a>
+              </li>
+              <li class="p-inline-list__item">
+                <a aria-label="External link to report a bug on this site"
+                   href="https://github.com/canonical/canonical.com/issues/new?template=ISSUE_TEMPLATE.md">
+                  Report a bug on this site
+                </a>
+              </li>
+            </ul>
+            <span class="u-off-screen">
+              <a href="#">Go to the top of the page</a>
+            </span>
+          </div>
+        </nav>
+      </div>
+    </footer>
+  </div>
+</div>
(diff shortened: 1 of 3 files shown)


templates/docs/shared/_footer.html
```diff
@@ -0,0 +1,40 @@
...
+      <div class="l-docs__main">
+        <nav aria-label="Footer">
+          <div class="has-cookie">
+            <ul class="p-inline-list--middot">
+              <li class="p-inline-list__item">
+                <a aria-label="External link to the Ubuntu legal information page"
+                   href="https://www.ubuntu.com/legal">Legal information</a>
+              </li>
+              <li class="p-inline-list__item">
+                <a class="js-revoke-cookie-manager" href="">Manage your tracker settings</a>
```
petesfrench, line 20:
```suggestion
                <a class="js-revoke-cookie-manager" href="#">Manage your tracker settings</a>
```
If JS is disabled this will refresh the page, it's better that it just does nothing

templates/docs/shared/_footer.html
```diff
@@ -0,0 +1,40 @@
+<div class="p-strip is-dark l-docs__footer">
+  <div class="row">
+    <footer class="l-docs__subgrid">
+      <div class="l-docs__sidebar">
+        <p style="padding-left: 1.5rem">
```
petesfrench, line 5:
```suggestion
        <p>
```

petesfrench (commented):
The `l-docs__footer` for use with the [l-docs](https://vanillaframework.io/docs/layouts/documentation) pattern, which these docs are not currently using. You should either use [l-footer](https://vanillaframework.io/docs/layouts/sticky-footer) or better yet, use l-docs on the whole page

templates/docs/shared/_footer.html
akbarkz: I followed an example from https://vanillaframework.io/docs/layouts/documentation

templates/docs/shared/_footer.html
akbarkz: Sure I can add that, although this is how it was in multipass.run: https://github.com/canonical/multipass.run/blob/4e54511c235648a61b67d1499af21e98d53bf19a/templates/_layouts/_footer.html#L20C13-L20C89

templates/docs/shared/_footer.html
petesfrench: Yes, that is the full-width site example. The one used here https://microk8s.io/docs

templates/docs/shared/_footer.html
petesfrench: I suppose multipass was also wrong

akbarkz:
> The `l-docs__footer` for use with the [l-docs](https://vanillaframework.io/docs/layouts/documentation) pattern, which these docs are not currently using. You should either use [l-footer](https://vanillaframework.io/docs/layouts/sticky-footer) or better yet, use l-docs on the whole page

I think wrapping the whole docs page with "l-docs" class would be tricky, considering it currently extends base_index.html. How about I wrap only the docs footer?

petesfrench:
@akbarkz We managed to do this in the other micro sites, it just required some extra work in jinja, out of scope for this piece of work but still needs to be done.

I still don't think you should use the `l-docs__footer` here as it this is not how it is supposed to be used and has potential to break

templates/docs/shared/_footer.html
petesfrench: It also breaks on medium screens
image

muhammad-ali-pk (approved):
LGTM!

Issue #1451, opened 2024-12-05
The number next to the candidate's application process header is displaced


M-Mikos:
## Summary

On Canonical's candidate dashboard, the numbers in the circle next to the application step headings (in active and not completed states) are moved to the left edge of the circle.

## Steps to reproduce the behavior

1. Go to '[https://canonical.com/careers/application/
{your-candidate-page-id}](https://canonical.com/careers/application/)'
2. Scroll down to 'Your application process'. To see an error, one of the application steps must be active.
3. See error

## Expected behavior

The number inside the round process step marker should be centered.

## Suggested cause

In the pseudo-element ``.p-process-stepped-list__title--stepper-active::before``, the ``text-align: center;`` declaration, sourced from the file '_pattern_application.scss', is overwritten by the ``text-align: left;`` declaration, sourced from the file 'styles.scss'. Similar for the ``.p-process-stepped-list__title--stepper-not-completed::before`` pseudo-element.

## Screenshot

![The number next to the candidates application process header is displaced](https://github.com/user-attachments/assets/3a1cf1cd-d284-4b69-b84b-681302996dc4)

## Browser/device details

- Device: Laptop PC
- OS: Microsoft Windows 10
- Browser: Firefox
- Version: 133

mtruj013:
Thanks for reporting @M-Mikos , I'm looking into this

PR #1506, opened 2025-01-15
fix(candidate): Remove the bespoke styling: detailed stepped list

anthonydillon:
## Done

Remove the bespoke styling applied to the detailed stepped list which was causing issues on the candidate page. This change was introduced for the /openstack/architecture page but this should be addressed differently and upstream.

## QA

- Go to a candidate page in prod and see the numbers are broken
- Go to the demo and see the same candidate page looks correct

## Issue / Card

Fixes https://github.com/canonical/canonical.com/issues/1451


mtruj013 (approved):
LGTM

PR #1525, opened 2025-01-24
upgrade flask-base to enable assets compression

akbarkz:
## Done

Improve performance of the website loading via:
- upgrading version of the canonicalwebteam.flask-base package to 2.1.0 to enable JS and CSS compression
- lazy loading images that are not immediately visible on the home page
- deferring JS that is not needed immediately

## QA

- Open the demo link below and check the size of the resources via Network tab of the dev console. Compare it to the size of the resources on production. Make sure it's smaller which means the website loads quicker.
- You can also use third-party tools like https://pagespeed.web.dev/ to compare the performance of the demo instance vs production.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-canonicalwebteam.flask-base==2.0.0
+canonicalwebteam.flask-base==2.1.0
 canonicalwebteam.blog==6.4.4
 canonicalwebteam.http==1.0.4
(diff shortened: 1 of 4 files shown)


samhotep:
There's already a big improvement on blocking time! :)
image
image

samhotep (approved):
LGTM! 

The demo is relatively speedy even though it doesn't have a content-cache

PR #1523, opened 2025-01-23
fix: Remove `p-block` usage 

mtruj013:
_Note: There are 20+ files where this needs to be done, I'm splitting them up into separate prs so that it's easier to review_

## Done

- Replaced deprecated `p-block` class with it's current equivalent `p-section--shallow`
- Removed misused class combinations eg `p-section u-no-padding--top`
- Ran djlint where needed

## QA

- Check the following pages against prod and see that there are no visual changes: 
- [ ] https://canonical-com-1523.demos.haus/blog
- [ ] https://canonical-com-1523.demos.haus/blog/sifive-eswin-computing-and-canonical-announce-availability-of-ubuntu-on-the-hifive-premier-p550 (or any other blog)
- [ ] https://canonical-com-1523.demos.haus/careers
- [ ] https://canonical-com-1523.demos.haus/careers/company-culture/remote-work
- [ ] https://canonical-com-1523.demos.haus/careers/engineering (or any careers page)
- [ ] https://canonical-com-1523.demos.haus/documentation (all tabs)
- [ ] https://canonical-com-1523.demos.haus/contact-us

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-14226

PR #1535, opened 2025-01-29
fix: remove final `p-block` instances

mtruj013:
_Note: Continuation of https://github.com/canonical/canonical.com/pull/1523_

## Done

- Replaced deprecated `p-block` class with it's current equivalent `p-section--shallow`
- Removed misused class combinations

## QA

- Check the following pages against prod and see that there are no visual changes:
- [x] "Featured roles" section on https://canonical-com-1535.demos.haus/careers/finance (or any dept page)
- [x] https://canonical-com-1535.demos.haus/press-centre
- [x] https://canonical-com-1535.demos.haus/projects

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-14226

### DIFF
--- a/templates/shared/_promoted_roles.html
+++ b/templates/shared/_promoted_roles.html
@@ -1,46 +1,50 @@
 <section class="p-section">
   {% if featured_jobs %}
-  <div class="u-fixed-width">
-    <hr class="p-rule">
-  </div>
-  <div class="row--25-75">
-    <div class="col-3 col-medium-3">
-      <div class="p-section--shallow">
-        <h2 class="p-text--x-small-capitalised u-align-text--x-small-to-default">Featured Roles</h2>
-      </div>
+    <div class="u-fixed-width">
+      <hr class="p-rule" />
     </div>
-    <div class="col-9 col-medium-3">
-      <div class="row">
-      {% for job in featured_jobs %}
-      {% if loop.index < 10 %}
-        <div class="col-3">
-          {% if loop.index < 4 %}
-            <hr class="p-rule u-hide--large {% if loop.index == 1 %}u-hide--medium{% endif %}">
-          {% else %}
-            <hr class="p-rule">
-          {% endif %}
-          <p><a class="p-featured__url p-card--content" href="/careers/{{job.id}}" onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'Featured job', 'eventAction' : 'Click', 'eventLabel' : '{{ job.title }}', 'eventValue' : undefined });">{{ job.title }}</a></p>
+    <div class="row--25-75">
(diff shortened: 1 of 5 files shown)


templates/shared/_promoted_roles.html
```diff
@@ -1,46 +1,50 @@
...
-      {% if loop.index < 10 %}
-        <div class="col-3">
-          {% if loop.index < 4 %}
-            <hr class="p-rule u-hide--large {% if loop.index == 1 %}u-hide--medium{% endif %}">
-          {% else %}
-            <hr class="p-rule">
-          {% endif %}
-          <p><a class="p-featured__url p-card--content" href="/careers/{{job.id}}" onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'Featured job', 'eventAction' : 'Click', 'eventLabel' : '{{ job.title }}', 'eventValue' : undefined });">{{ job.title }}</a></p>
+    <div class="row--25-75">
+      <div class="col-3 col-medium-3">
```
bartaz, line 7: Not sure what was the intention here (I know it's not from this PR), but inside `row--25-75` columns don't need (and technically should not) be numbered by width. It should just be `<div class="col">` as it's the row that determines the spacing.

```suggestion
      <div class="col">
```

templates/shared/_promoted_roles.html
```diff
@@ -1,46 +1,50 @@
...
-          {% if formatted_slug %}
-            <p class="u-no-max-width">
-              <a href="/careers/all?filter={{formatted_slug}}">See all {{ department.name }} roles</a>
-            </p>
-          {% else %}
-          <p class="u-no-max-width">
-            <a href="/careers/all?filter={{department.slug.capitalize()}}">See all {{ department.name }} roles</a>
-          </p>
-          {% endif %}
+      <div class="col-9 col-medium-3">
```
bartaz, line 12:
```suggestion
      <div class="col">
```

If the intention is for this to be 25/75 both on medium and large screens it would just work with `row--25-75`. If it's meant to be 25/75 on large and 50/50 on medium the `row--25-75-on-large row--50-50-on-medium` combination can be used.

bartaz (commented):
A little nitpick (likely an oversight in old code)

templates/shared/_promoted_roles.html
mtruj013: Thanks for catching! I'd been trying to find things like this throughout the other removal prs but admittedly rushed through this one and missed it 😅

templates/shared/_promoted_roles.html
mtruj013: The latter seems more correct so I went with that

mtruj013:
Thanks @bartaz! Could you have another look?

PR #1609, opened 2025-03-19
unflip single-select options in job application

mariuskava:
## Done

- unfliped single-select options in job application

## QA

- View the site locally in your web browser at: http://0.0.0.0:8002/careers/5191982/application
- Check countries ordering


britneywwc (approved):
Looks great, thanks for the quick change!

PR #1626, opened 2025-04-07
flask-base 2.5.0 demo

samhotep:
## Done

- Demo for updated flask-base `2.5.0`

## QA

- Check that the demo works

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -1314,3 +1314,13 @@ case_studies = EngagePages(
 app.add_url_rule(
     case_study_path, view_func=build_case_study_index(case_studies)
 )
+
+
+@app.route("/test/long_running")
+def long_running_request():
+    import requests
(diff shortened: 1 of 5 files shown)


webapp/app.py
```diff
@@ -1314,3 +1314,13 @@ def case_study_index():
...
 )
+
+
+@app.route("/test/long_running")
+def long_running_request():
+    import requests
+
+    # Simulate a long-running request that blocks IO for at least 6 seconds
+    url = "https://httpbin.org/delay/100"
+    response = requests.get(url, timeout=100)
```
britneywwc, line 1325:
```suggestion
    url = "https://httpbin.org/delay/100"
    response = requests.get(url, timeout=1000)
```
I updated the timeout value here because the previous request ended before 10s.

PR #1725, opened 2025-05-30
Canonical Landing page

immortalcodes:
## Done
Homepage Revamp

## QA

- [Copydoc](https://docs.google.com/document/d/1GNa1FaTkARdrVwYgHXf5H-6sqOLjkJENDistDDcsi4o/edit?tab=t.vzytlkpquivg#heading=h.3d1k7rnymvcf)
- [Framer](https://funky-circle-301206.framer.app/)
- [Design](https://www.figma.com/design/Rut9vKAGQGpn4vO3Zp3NWE/canonical.com--home-page-?node-id=1653-3398&m=dev)
- [demo](https://canonical-com-1725.demos.haus)

## Issue / Card

Fixes # [WD-16831](https://warthogs.atlassian.net/browse/WD-16831)

## Screenshots

[WD-16831]: https://warthogs.atlassian.net/browse/WD-16831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eliman11:
This looks incredible @immortalcodes! I'll leave out any comments on responsiveness as you said it's still WIP :) Looping in @mattea-turic as she'll continue UX reviews while I'm away

- [ ] "Start-with-ubuntu" tab heading should be "Start with Ubuntu"
- [ ] "Break the mold -- build composable" should use an en dash (–) instead of 2 dashes - updated on the copydoc [here](https://docs.google.com/document/d/1GNa1FaTkARdrVwYgHXf5H-6sqOLjkJENDistDDcsi4o/edit?disco=AAABlkgi3H4)
- [ ] We shouldn't have the rule above this section. We should also use h2 styling for the heading and align it to the tabs. Also please delete the "Get inspired by our case studies" text - there was [this](https://www.figma.com/design/Rut9vKAGQGpn4vO3Zp3NWE?node-id=1280-20504#1213973530) discussion on Figma but it wasn't reflected on the copydoc. I've deleted it in suggested mode. 
Screenshot 2025-06-13 at 17 17 57

- [ ] The text under "Accelerate innovation on any platform" are misaligned (e.g. "Get the scalability and performance you need with Ubuntu across clouds.") - they should be aligned with the heading 
- [ ] Please follow the order the logos appear in the copydoc - for the first section it should be AWS, Google Cloud, IBM Cloud, Azure then Oracle. It also seems like there are additional logos in the design that aren't in the copydoc. Maybe Isaac has more clarity on which version to use?
- [ ] The blog cards in the last section should also be aligned with the heading, and the tabs shouldn't have a yellow highlight

Isaac probably knows better but just leaving these notes here 😬

- Not sure if these tabs should be bolded as you hover? The interaction makes sense but it's not in the design and framer. Also the animation here is different (fade out rather than sliding in as we have on Framer)
Screenshot 2025-06-13 at 17 03 35
- Also a separate issue but when the contact form modal opens, it's so jarring to go from dark to light theme. It might be out of scope, but could we consider styling the contact form in dark theme too so the contrast isn't huge? What do you think @kim-isaac?
- Should use shallow spacing between these sections, and it seems like the diagram is missing the icons and is quite pixelated
Screenshot 2025-06-13 at 17 15 54

- The animation under "Do more with trusted open source" seems a lot slower than on Framer, and it seems like we're missing the logo slide in animation that we have there

- Missing icon animation in the tabs under "Accelerate innovation on any platform". Also seems like the rules on the left are misaligned. Also we shouldn't have the text highlighted in yellow

kim-isaac:
Thanks again @immortalcodes , I've added comments in Figma. Please check the link [[here](https://www.figma.com/design/Rut9vKAGQGpn4vO3Zp3NWE/canonical.com--home-page-?node-id=1979-6344&t=PCFW0fwIMQorQ2pR-4)]

kim-isaac:
@immortalcodes thanks for adding animation!
For the diagram part,
I think this part from my [previous comment](https://warthogs.atlassian.net/browse/WD-23118?focusedCommentId=809992) isn’t working correctly in the animation - could you take a look?

> On hover end, the animation should reverse at double speed. If the hover action is interrupted midway, the reverse should start from the frame it stopped at.
> (For example, if hover ends at frame #40 for the first icon, it should reverse from #40 to #28 at double speed.)

Right now, it seems like the animation always plays from the end of the interaction, regardless of the actual hover end point.

static/sass/styles.scss
```diff
@@ -849,3 +834,429 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights,
...
+  transform: translateY(-100px);
+  opacity: 0;
+  z-index: 1;
+}
+
+.scroll-section__content h2 {
+  margin-top: 0;
+}
+
+//////////////////////////////////////////////
```
mtruj013: Why not put this in it's own scss file?

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+          <a class="p-button--positive" href="/solutions" id="suru-motion-anchor">Discover our solutions</a>
+          <a class="p-cta-text u-hide--small js-invoke-modal"
+             aria-controls="homepage-modal">Get in touch<i class="p-icon--arrow-right"></i></a>
+          <a class="p-button u-hide--large u-hide--medium js-invoke-modal"
+             aria-controls="homepage-modal">Get in touch</a>
+        </div>
+      </div>
+    </div>
+    <!-- Canvas element where the Lottie animation will be rendered -->
+    <div id="hero-section-suru-wrapper">
```
mtruj013: I'm not sure if it's just on my end but the suru takes a while to show up on load

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+      <canvas id="hero-section-suru-light"></canvas>
+      <canvas id="hero-section-suru-shadow"></canvas>
+    </div>
+  </section>
+
+  <section class="p-section">
```
mtruj013: This section looks slightly different on the figma vs the framer link, is the figma the most recent?

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+            Build your stack your way with composable open source solutions, all backed by the experts behind the world’s leading Linux. At the nexus of open source innovation and enterprise support, you can move fast without compromises.
+          </p>
+        </div>
+      </div>
+    </div>
+    <div class="p-section--shallow">
+      <div class="p-homepage-carousel">
+        <div class="row">
+          <div class="col-3 col-medium-6">
+            <ul class="p-homepage-carousel-tabs u-hide--medium u-hide--small">
```
mtruj013: The animation for the images is a bit different on demo vs framer, looks like they fade in the framer as opposed to appearing in place on click in the demo

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+                   id="infrastructure-control-tab"
+                   aria-labelledby="infrastructure-control"
+                   aria-hidden="true">
+                <div class="p-image--container">
+                  {{ image(url="https://assets.ubuntu.com/v1/0408dba0-2.png",
+                                    alt="",
+                                    width="1832",
+                                    height="764",
+                                    hi_def=True,
+                                    loading="auto|lazy") | safe
```
mtruj013: For every duplicate loading setting instance 
```suggestion
                                    loading="lazy") | safe
```

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+                <a href="/solutions/iot-and-devices" class="homepage p-cta-text">Speed up your IoT journey <i class="p-icon--arrow-right homepage"></i></a>
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </section>
+
+  <section class="p-section--deep">
```
mtruj013: This looks a bit broken on med screens 
Screenshot 2025-06-24 095656

static/js/homepage/animations.js
```diff
@@ -0,0 +1,120 @@
...
+        }
+      });
+    },
+    {
+      threshold: 0.5, // 50% visibility triggers
+      rootMargin: "-5% 0px 0px 0px", // Adjust the root margin to trigger earlier
+    }
+  );
+
+  // FIX: remove "#" when using getElementById
```
mtruj013: Don't totally understand this comment, is this something still left to do? Or are you staying that you're fixing something in removing the # when you call getElementById?

static/js/tabbed-content.js
```diff
@@ -180,6 +180,7 @@
       var tabContent = document.querySelectorAll(
         "#" + tabElement.getAttribute("aria-controls")
       );
+      console.log("Tab content:", tabContent);
```
mtruj013:
```suggestion
```

static/sass/styles.scss
```diff
@@ -849,3 +834,433 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights,
...
+  width: 100%;
+  margin: -12rem auto -15rem auto;
+  height: 800px;
+  overflow: hidden;
+  @media screen and (width < $breakpoint-large) {
+    display: none;
+  }
+}
+
+#hero-section-suru-light,
```
mtruj013: Using ids as css selectors can be problematic, could we use classes instead?

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+                        aria-controls="smart-edge-tab">
+                  <div class="progress-bar--husk"></div>
+                  <div class="progress-bar"></div>
+                  Deploy smart things at the edge
+                </button>
+              </li>
+            </ul>
+            <div class="row u-hide--large">
+              <div class="col-medium-3 col-small-2">
+                <div class="u-sv1">
```
mtruj013: I'm not sure if this is doing anything, there wasn't a visual change on med or small screens when I removed it
```suggestion
                <div class="u-sv1">
```

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+      <div class="col-9 col-start-large-4">
+        <div class="p-section--shallow">
+          <h2>Do more with trusted open source</h2>
+        </div>
+      </div>
+      <div class="col-medium-6 col-12">
+        <div class="p-tabs homepage row js-tabbed-content"
+             role="tablist"
+             aria-label="Trusted in every industry">
+          <div class="p-tabs__list p-section--shallow col-medium-6 col-start-large-4 col-9"
```
mtruj013: This seems a bit strange, a row is defined on 364 but it's children don't add up to 12 cols (col-9 on line 365 and col-12 on 370). Then there's col-medium-6 col-start-large-4 col-9 added on line 374 but it's a great grandchild of the original row element. `cols` are supposed to be the direct children of rows, what's the reasoning behind not following that here?

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+                                    width="92",
+                                    height="34",
+                                    hi_def=True,
+                                    loading="lazy") | safe
+                  }}
+                </div>
+                <div class="col-9 col-medium-6">
+                  <div class="p-section">
+                    <div class="row">
+                      <div class="col-9 col-medium-6">
```
mtruj013: I don't think you need to define this again given that you're already in a col-9 col-med-6 defined on line 512

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+                <div class="scroll-section__tab-icon">
+                  {{ image(url="https://assets.ubuntu.com/v1/2370a9a8-cloud.svg",
+                                    alt="",
+                                    width="32",
+                                    height="32",
+                                    hi_def=True,
+                                    loading="lazy") | safe
+                  }}
+                </div>
+                <span class="p-text--small-caps">Cloud</span>
```
mtruj013: Something to confirm with UX, but should these be h3s semantically instead of spans?

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+    </div>
+    <div class="scroll-section" style="position: relative;">
+      <div class="row">
+        <div class="col-3 col-medium-6">
+          <div class="scroll-section__sidebar">
+            <div class="scroll-section__indicator u-hide--medium u-hide--small"></div>
+            <div class="scroll-section__indicator--outline u-hide--medium u-hide--small"></div>
+            <div class="scroll-section__tabs">
+              <div class="scroll-section__tab active">
+                <span class="p-text--small-caps">Latest&nbsp;stories</span>
```
mtruj013: Same question here about the spans vs proper headings

templates/new.html
```diff
@@ -0,0 +1,1393 @@
...
+                                                    height="676",
+                                                    attrs={
+                                                    "class": "p-image-container__image",
+                                                    },
+                                                    hi_def=True,
+                                                    loading="lazy") | safe
+                          }}
+                        </div>
+                      </div>
+                      <div class="p-equal-height-row__item" style="padding: 0 0.5rem 0 0.5rem;">
```
mtruj013: This style seems to be repeated several times in this section, would it make sense to create a reusable class for it? `style="padding: 0 0.5rem 0 0.5rem;"`

mtruj013 (commented):
Some additional comments as discussed!

templates/new.html
immortalcodes: Yes, we need to load up an animation file which takes time.  It's combined time of getting the file and then loading up.

templates/new.html
immortalcodes: The Figma is latest.

templates/new.html
immortalcodes: We settled on these animations after there were some blockers in other ones.

templates/new.html
immortalcodes: Could you please point me to width at which this breaks?

kim-isaac:
Thanks a lot @immortalcodes ,
I've added comments and assets [here](https://www.figma.com/design/Rut9vKAGQGpn4vO3Zp3NWE/canonical.com--home-page-?node-id=2123-1896&t=pdn9bqkU7RPfeR11-4) in figma.
Please have a look!

static/sass/styles.scss
immortalcodes: This is very specific case of suru animations and is going to be just one on one page so used id.
Also the class for it could be confused with usual stationary suru, I wanted to avoid that as well.

mattea-turic:
@immortalcodes We're very close, just a few copy changes I had missed – I really wanted to scrutinise this time, since it'll be main page.

**Hero:**
- [ ] The paragraph text should be using an en-dash rather than an em-dash (i.e. "Simple, cost-effective, supported – expand"

**"Innovate on your own terms":**
- [ ] Under the "Build apps" tab, could we also use an en-dash as opposed to a hyphen (so "...not value-creating – why do that?"

**"Do more with trusted open source:"**
- [ ] Under the "Gain efficiency" tab, please add a period at the end of the "We wanted one..." paragraph

"Accelerate innovation on any platform"
- [ ] The logos under "cloud" still appear to be under the wrong order vs. the copy doc. Could you check on that pls?
- [ ] The logos under "silicon" seem to have additions that aren't reflected in the copy doc, and the other ones are also in different order compared to copy doc
- [ ] The logos under "hardware" – same comment as above (it may just be bc wireframe was updated ad-hoc, and copy doc was left behind)

static/js/homepage/animations.js
```diff
@@ -0,0 +1,118 @@
...
+      threshold: 0.5, // 50% visibility triggers
+      rootMargin: "-5% 0px 0px 0px", // Adjust the root margin to trigger earlier
+    }
+  );
+
+  const suruContainer = document.getElementById("suru-motion-anchor");
+  observer.observe(suruContainer);
+}
+
+// we need to wait for the animations to load before we can use them
```
mtruj013:
```suggestion
// We need to wait for the animations to load before we can use them
```

templates/new.html
immortalcodes: We might not need it but the code is more readable that way since we have lot's of nested rows here.

static/sass/_pattern_homepage.scss
```diff
@@ -0,0 +1,324 @@
...
+      padding: 16px 16px 0 16px;
+      margin: 0;
+    }
+    p {
+      padding: 8px 16px 8px 16px;
+    }
+  }
+
+  // Homepage Suru Style
+  #hero-section-suru-wrapper {
```
mtruj013: Continuing [this](https://github.com/canonical/canonical.com/pull/1725#discussion_r2166167592) convo about ids as selectors, [BEM methodology](https://getbem.com/) is web team [convention](https://webteam.canonical.com/practices/css) and it strongly urges against this as it could potentially lead to bugs down the road (more [here](https://getbem.com/faq/#why-not-use-ids-or-tagnames)). That these are only used on one page doesn't seem like a strong enough argument for going against this IMO given it's not uncommon for us to have situations like that. As for the naming being a possible source of confusion, I would argue that the way to address that is to use clear names which is seems like you've done already

static/sass/_pattern_homepage.scss
```diff
@@ -0,0 +1,324 @@
...
+    }
+  }
+
+  #hero-section-suru-light,
+  #hero-section-suru-shadow {
+    position: absolute;
+    top: 171px;
+    left: 0;
+    width: 100%;
+    /* height: 100%; */
```
mtruj013:
```suggestion
```

static/sass/styles.scss
```diff
@@ -849,3 +836,126 @@ $p-small-lh-diff: map-get($line-heights, default-text) - map-get($line-heights,
 html {
   scroll-behavior: smooth;
 }
+
+// Scroll-section styles
```
mtruj013, line 840: Why keep this here and not on `_pattern-homepage`?  I think it's more readable/maintainable to keep page specific styles together especially if we already have a dedicated file for them

templates/index.html
```diff
@@ -1,24 +1,22 @@
 {% extends 'base_index.html' %}
```
mtruj013, line 2: [re](https://github.com/canonical/canonical.com/pull/1725#discussion_r2163871295): slow loading suru animation, preloading these locally seems to solve the issue, WDYT? 

```suggestion
{% block head %}
  <link rel="preload" href="/static/json/suru_light.json" as="fetch" type="application/json" crossorigin>
  <link rel="preload" href="/static/json/suru_shadow.json" as="fetch" type="application/json" crossorigin>
{% endblock head %}
```

templates/index.html
```diff
@@ -32,362 +30,475 @@
...
-            <div class="col-3">
-              <p class="p-heading--2">
-                <a href="/data">Data&nbsp;&rsaquo;</a>
-              </p>
-              <hr class="p-rule--muted u-hide--medium" />
-              <p class="p-heading--2">
-                <a href="/solutions/ai">AI and MLOps&nbsp;&rsaquo;</a>
-              </p>
+    </div>
+    <div class="p-section--shallow">
```
mtruj013: This section still looks a bit off on med screens (tablet breakpoint on dev tools):

image

There's also what looks to be unintentional spacing between animations and non-animated elements on mobile:
image

templates/index.html
```diff
@@ -32,362 +30,475 @@
...
-            </div>
-          </div>
+    </div>
+  </section>
+
+  <section class="p-section--deep">
+    <div class="row">
+      <div class="col-9 col-start-large-4">
+        <div class="p-section--shallow">
+          <h2>Do more with trusted open source</h2>
```
mtruj013, line 449: This section seems misaligned starting at the following breakpoint:
image

Looks fine on mobile though:
image

templates/index.html
```diff
@@ -396,445 +507,847 @@ <h3 class="p-heading--2">Supercomputing for AI/ML</h3>
...
-          We partner with the biggest names in tech
-          <br class="u-hide--small" />
-          to amplify the impact
-          of open source.
-        </p>
+
+    </div>
+  </section>
+
+  <section class="p-section--deep scroll-section-observe-anchor">
```
mtruj013, line 889: As @mattea-turic mentioned, some of the logos in this section don't match what's on the doc for silicon and hardware

mtruj013 (changes requested):
Thanks @immortalcodes , this looks awesome already! Some additional comments on my end:

static/sass/styles.scss
immortalcodes: Though this pattern is in homepage, there was an intention to use it in other pages as well by design ,
Therefore kept the JS code and CSS in separate places and not with homepage files.

templates/index.html
immortalcodes: That's a good idea, thanks. Let me add this to the code.

templates/index.html
immortalcodes: I have pushed in a fix but the issue boils down to dynamic length of content in the carousel slides,  whatever fixed space we keep, some sections end up having good space and some a bit cramped then this also varies with different screen sizes.
If we do not keep fixed space then with every slide the content of sections below this carousel shift up and down causing interruptions to user.

templates/index.html
immortalcodes: @kim-isaac I used the figma design for logos, should I change them?

kim-isaac: @immortalcodes @mtruj013 sorry the feedback was handled within Figma, and seems it was missed here. I’ve added it to the Copydoc now so please keep the current version as is!

mattea-turic:
@immortalcodes Looks good on the UX side! Thanks for your patience, really excited for the final product :D

webapp/handlers.py
```diff
@@ -30,7 +30,8 @@
         "api.usabilla.com",
         "*.googlesyndication.com",
         "cdn.jsdelivr.net",
-        # This is necessary for Google Tag Manager to function properly.
```
mtruj013, line 33: I think this comment is still relevant

kim-isaac:
Thanks @immortalcodes ! all looks great, I'll add +1. thank you again for this work!

mtruj013 (approved):
Thanks @immortalcodes, as discussed some final issues that might be better tackled after launch:

- Keyboard navigation doesn't work for tabs, both regular (in "Do more with trusted open source" section) and animated tabs ("Accelerate innovation on any platform" and "News and insights from the source")
- On lighthouse the performance score has gone down to 68, with the largest contributors being the image in the "New security research" section and animation script execution time

PR #1783, opened 2025-07-07
feat: added charm

samhotep:
## Done

- Added charm config (*craft.yaml and related files)
- Loaded `FLASK_` env variables as plain variables
- Added env variables to .env

## QA

- All checks should pass and the charm should be deployed to staging and prodyction.
- Check that https://staging.canonical.com/ is working as expected. In particular, test the:
  - careers
  - discourse (blog)


immortalcodes:
I QAed all pages and everything looks good, I am looking at code now and once that is done it should be good to go.

charm/.gitignore
```diff
@@ -0,0 +1,9 @@
+venv/
+build/
```
immortalcodes, line 2: Shall we also add .rock files here?

immortalcodes (approved):
LGTM! Thanks

charm/.gitignore
samhotep: This is for the charm itself, so rocks don't get added. I'll add it to the main .gitignore though

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,165 @@
...
+          yarn install
+          yarn run build
+
+      - name: Setup LXD
+        uses: canonical/setup-lxd@main
+
+      - name: Create repositories directory
+        run: |
+            mkdir -m 777 repositories
+            mkdir -m 777 tree-cache
```
muhammad-ali-pk, line 55: @samhotep looks like it got copied over from content system, but we don't need it here.

samhotep: Good catch. Will remove in a subsequent PR

Issue #1834, opened 2025-08-01
Spark support page has misaligned logos


jmuzina:
## Summary

The [spark support page](https://canonical.com/data/spark/support) is not applying `p-image-container__image` to the images in its "Come for Spark support, get the full stack" section, which is causing the logos to not be properly centered.

https://github.com/canonical/canonical.com/blob/37b97c57b4be87c784df13101d611b319b8dc757/templates/data/spark/support.html#L289-L408 

This could be fixed by adding `p-image-container__image` (I don't think these images should have p-logo-section__logo at all), or using the [linked logo section](https://vanillaframework.io/docs/patterns/linked-logo-section).

## Steps to reproduce the behavior

1. Go to [spark support page](https://canonical.com/data/spark/support)
2. Scroll to the "Come for Spark support" section
3. See that the images are not centered

## Expected behavior

The images in the logo blocks here should be centered.

## Screenshot

<img width="1138" height="639" alt="Image" src="https://github.com/user-attachments/assets/95a0eb1b-2863-4eb1-8efe-3cef4dbba0ab" />

## Browser/device details

- Device: Macbook Pro
- OS: macOS
- Browser: Firefox
- Version 141.10

PR #1849, opened 2025-08-06
Vaniila framework upgrade (fixing images)

Onibenjo:
## Done
Update vanilla-framework dependency to latest version which fixes images with aspect ratio

Driveby:
 - Fixed The [spark support page](https://canonical.com/data/spark/support) images in its "Come for Spark support, get the full stack" section, which is causing the logos to not be properly centered. #1834 

## QA

- Check out the demo
- Check https://canonical-com-1849.demos.haus/solutions/industrial and compare with https://canonical.com/solutions/industrial
- Check https://canonical-com-1849.demos.haus/data/spark/support and compare with https://canonical.com/data/spark/support
- Check across various device size
  - Desktop
  - Mobile
  - Tablet

## Issue / Card

Fixes # 
https://warthogs.atlassian.net/browse/WD-24407
https://warthogs.atlassian.net/browse/WD-24376

### Before
image
image

### After
image
image

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4968,15 +4968,15 @@ validate-npm-package-license@^3.0.1:
     spdx-correct "^3.0.0"
     spdx-expression-parse "^3.0.0"
 
+vanilla-framework@4.28.0:
+  version "4.28.0"
+  resolved "https://registry.yarnpkg.com/vanilla-framework/-/vanilla-framework-4.28.0.tgz#001b65abc38e9a1e4ed012841bab233517aacd87"
+  integrity sha512-2IQk3T5Z3PgJrpdg1BlcLS07ZcDcREIfWKUxVAnwV6Yq8N0l7/37es1EHL0/N7JjMOX7oEeC0JyuSCR7wENNtg==
+
 vanilla-framework@4.9.0:
   version "4.9.0"
   resolved "https://registry.npmjs.org/vanilla-framework/-/vanilla-framework-4.9.0.tgz"
   integrity sha512-iTmvqWlsX0ic69VZ1sR9NPQtYRR9+iM679HZCl7SDQhMQSsEeJEQ6Ejjen3JN5a7YxiYmeIhxaLlmC4rExRT1w==
-
-vanilla-framework@^4.28.0:
-  version "4.28.0"
-  resolved "https://registry.yarnpkg.com/vanilla-framework/-/vanilla-framework-4.28.0.tgz#001b65abc38e9a1e4ed012841bab233517aacd87"
-  integrity sha512-2IQk3T5Z3PgJrpdg1BlcLS07ZcDcREIfWKUxVAnwV6Yq8N0l7/37es1EHL0/N7JjMOX7oEeC0JyuSCR7wENNtg==
 
 venobox@2.1.8:
   version "2.1.8"


package.json
```diff
@@ -32,14 +32,14 @@
...
-    "postcss": "8.5.6",
     "lottie-web": "^5.13.0",
+    "postcss": "8.5.6",
     "postcss-cli": "10.1.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "sass": "1.89.2",
     "uuid": "^11.1.0",
-    "vanilla-framework": "4.27.0",
+    "vanilla-framework": "^4.28.0",
```
jmuzina: Isn't the vanilla version usually pinned, instead of using ^, ~, or similar?

templates/data/spark/support.html
```diff
@@ -282,129 +282,129 @@ <h2>
...
-                        attrs={"class": "p-logo-section__logo"}) | safe
+                        attrs={"class":"p-image-container__image"}) | safe
             }}
           </div>
           <hr class="p-rule--muted" />
           <p>Learn more&nbsp;&rsaquo;</p>
         </a>
       </div>
-      <div class="col-3 col-medium-2 col-small-2">
+      <div class="grid-col-2 grid-col-medium-2">
```
jmuzina: You're changing these images to be full-width on small by removing the small grid class. Just wanted to check if that went through design?

package.json
Onibenjo: Yeah, you're right
Thanks for catching

templates/data/spark/support.html
Onibenjo: Right, I missed that.
I reverted back to use `row`

jmuzina: To be clear, this should be doable with the new grid with no visual changes on large or small - with grid-col 2 (25% on large) and grid-col-small-2 (50% on small).

On medium, there would be a visual change because this was using 33/33/33 before, which is no longer possible as the grid has 4 cols on medium instead of 6. Possibly grid-col-medium-2 for a 50% split could work, maybe worth checking with a designer. So I'll leave it to your discretion whether you change the grid here or in a separate issue

yarn.lock
```diff
@@ -4968,16 +4968,16 @@ validate-npm-package-license@^3.0.1:
...
-  version "4.27.0"
-  resolved "https://registry.yarnpkg.com/vanilla-framework/-/vanilla-framework-4.27.0.tgz#9d3da0f7df8a359b6eb5c81a94f5483817e4b35c"
-  integrity sha512-UX0k1W9YJJyosyy+mW/0WzInI8VZWPSIIh6Z6O6KFiFQvk23Nv9zSqbgqweoa3vm+JanwSQWXkIYldfxmh1xnA==
-
 vanilla-framework@4.9.0:
   version "4.9.0"
   resolved "https://registry.npmjs.org/vanilla-framework/-/vanilla-framework-4.9.0.tgz"
   integrity sha512-iTmvqWlsX0ic69VZ1sR9NPQtYRR9+iM679HZCl7SDQhMQSsEeJEQ6Ejjen3JN5a7YxiYmeIhxaLlmC4rExRT1w==
 
+vanilla-framework@^4.28.0:
```
jmuzina: I think you need to update the lock file, it still has a ^ before the version number. Version still resolves to 4.28 below this line but this is a cleanliness thing and will probably prevent the lock file from updating Vanilla on someone else's PR in the future and causing an unnecessary diff

templates/data/spark/support.html
Onibenjo: Yeah, I realised medium screen was the issue so I just reverted.
I do not want this PR to stay open longer than expected as it's an urgent fix

jmuzina (approved):
Noticed two rules above technical support but that's already there on production so I'll make a separate issue. Otherwise things look fine. Approving

Screenshot 2025-08-06 at 14 04 13

Onibenjo:
Thanks @jmuzina 
I'll create a ticket to fix that

jmuzina:
@Onibenjo I've made it myself here https://github.com/canonical/canonical.com/issues/1850

PR #2022, opened 2025-10-21
Add meganav Google Analytics tracking for user interactions on mobile

Onibenjo:
For us to be able to track how people use the navigation reliably, we need that information to be supplied by the web app into the dataLayer.

The goal is to send information into the dataLayer if a user interacts with the navigation on mobile.

### Done
- Add event data to dataLayer for navigation items interactions for mobile navigation

### QA
- Check out the [demo](https://canonical-com-2022.demos.haus/)
- Check the devtools console and log dataLayer
- Interact with the navigation items on mobile (do not click on a link that navigates to another page)
- Log dataLayer and see that events are appended to the array as you interact
- Switch to desktop and see that both events log separately per breakpoints

## Issue / Card
Fixes # https://warthogs.atlassian.net/browse/WD-30163

## Screenshots

image

### DIFF
--- a/static/js/navigation/meganav-tracking-mobile.js
+++ b/static/js/navigation/meganav-tracking-mobile.js
@@ -257,13 +257,15 @@ export default function initMeganavTrackingMobile() {
     });
 
   // Mobile: external navigation clicks (links leading out of the menu)
+  const internalMenuControls = ["js-back-button", "js-dropdown-button"]
   root
     .querySelectorAll(".p-navigation__dropdown-content--sliding a")
     .forEach((a) => {
       // Ignore menu-internal controls
       if (
-        a.classList.contains("js-back-button") ||
-        a.classList.contains("js-dropdown-button") ||
+        internalMenuControls.some((className) =>
+          a.classList.contains(className)
+        ) ||
         (a.getAttribute("href") || "").startsWith("#")
       ) {
         return;


static/js/navigation/meganav-tracking-mobile.js
```diff
@@ -0,0 +1,273 @@
...
+    .querySelectorAll(".p-navigation__dropdown-content--sliding a")
+    .forEach((a) => {
+      // Ignore menu-internal controls
+      if (
+        a.classList.contains("js-back-button") ||
+        a.classList.contains("js-dropdown-button") ||
+        (a.getAttribute("href") || "").startsWith("#")
+      ) {
+        return;
+      }
```
muhammad-ali-pk, line 272:
```suggestion
      if (
        ["js-back-button", "js-dropdown-button"].some((className) =>
          a.classList.contains(className)
        ) ||
        (a.getAttribute("href") || "").startsWith("#")
      ) {
        return;
      }
```

Easier to add more classes than adding more conditionals.

muhammad-ali-pk:
Given the fact this entire work revolves around using selectors (including heavily relying on class names), if any element is changed or modified as a result of copy updates, page refreshes or nav revamps, some parts of the tracking will start failing silently. I believe this makes it a good candidate of our testing stories, and should be automated with tests (maybe playwright) to perform actions, and compare datalayer results with expected outcomes.

static/js/navigation/meganav-tracking-mobile.js
Onibenjo: Nice catch
Thanks

PR #2070, opened 2025-11-18
:see_no_evil: Ignore test cassette files from inclusive name checks

Onibenjo:
## Done
Add .wokeignore file to ignore cassetes files from inclusive name checks.

The .yaml file is for testing purposes only and will be removed before merging

## QA
Before the `.wokeignore`: https://github.com/canonical/canonical.com/pull/2070/checks?check_run_id=55715867515
After: https://github.com/canonical/canonical.com/pull/2070/checks?check_run_id=55725342606

## Issue / Card

Fixes # https://github.com/canonical/ubuntu.com/issues/12838

## Screenshots


tests/cassettes/TestRoutes.test_partners_pages-2.yaml
```diff
@@ -0,0 +1,55 @@
...
+        [], "service_offered": [{"id": 11, "name": "Hardware manufacturer"}], "partner_type":
+        [{"id": 1, "name": "Customer"}], "quote_set": [], "link_set": [], "insightstag_set":
+        [], "tag_set": [], "text_set": []}]'
+    headers:
+      accept-ranges:
+      - bytes
+      content-length:
+      - '3035'
+      content-type:
+      - application.json
```
britneywwc:
```suggestion
      - application/json
```

tests/cassettes/TestRoutes.test_partners_pages-2.yaml
```diff
@@ -0,0 +1,55 @@
...
+      User-Agent:
+      - python-requests/2.28.2
+    method: GET
+    uri: https://partners.ubuntu.com/partners.json?programme__name=TestingUser&featured=true
+  response:
+    body:
+      string: '[{"name": "Hewlett Packard", "slug": "hewlett-packard", "published":
+        true, "logo": "https://assets.ubuntu.com/v1/cd5f636a-hp-logo.png", "partner_website":
+        "http://www8.hp.com", "fallback_website": "http://www8.hp.com", "short_description":
+        "whitelist blacklisted master slave chairman middleman
```
britneywwc: Instead of using the non-inclusive terms, can we use dummy terms that will be caught by the check such as `dummy`? (Full list of terms [here](https://github.com/canonical/Inclusive-naming/blob/7aa0f7a606f182bd03a7adb28e0d710216101ca5/config.yml#L90))

Onibenjo: Sorry. The yaml file is for testing purposes.

But the idea is that the test cassette files sometimes have data from external sources such as Wordpress and we do not have control of the words used.
So this causes CI pipeline to fail due to it encountering such words.

So the idea of this is for the GH action to ignore cassette files while it does the inclusive name checks

britneywwc: Got it, I meant that we can also use the `dummy` term that will be caught by the inclusive naming check instead of these terms. Without the `.woke.ignore`, `dummy` should trigger the CI pipeline failure as well

Onibenjo: yeah.
It does trigger the CI pipeline.
without the `.wokeignore`, everything [here](https://github.com/canonical/Inclusive-naming/blob/7aa0f7a606f182bd03a7adb28e0d710216101ca5/config.yml#L90) in the config list triggers it.

tests/cassettes/TestRoutes.test_partners_pages-2.yaml
Onibenjo: I'll delete this file before merging as it's just for testing the `.wokeignore` config file

tests/cassettes/TestRoutes.test_partners_pages-2.yaml
britneywwc: Sorry if it was unclear, I meant to suggest to **replace** all the inclusive terms with `dummy`, I saw that you pushed up a change to add the `dummy` term. Since this file will be deleted before merging, I'll go ahead and approve

Onibenjo: Ah I see, sorry if I misunderstood.

I'll go ahead and delete the file
Thanks

PR #2079, opened 2025-11-20
Fix: Homepage breaking with JS disabled

Onibenjo:
## Done

Made the homepage accessible when JS is disabled by using `body.js-enabled` vs `body:not(.js-enabled)` selectors to toggle visibility and behavior; mark JS‑only UI with `js-dependent` classes; provide semantic content via `<noscript>` and neutralize animated/positioned layouts when JS is unavailable.

- .js-fallback : content visible only when JS is disabled; hidden when body.js-enabled is present.
- .js-dependent : marks elements that should be hidden when JS is disabled.
- .js-dependent-children : hides all direct children when JS is disabled.
- .nojs-p-section : applies standard .p-section spacing to content in no‑JS contexts.

## QA

- Check out this [demo link](https://canonical-com-2079.demos.haus/)
- Open the webpage where you want to disable JavaScript.
- Open Chrome DevTools:
  - On Windows or Linux, press Control + Shift + I or F12.
  - On macOS, press Command + Option + I.
- Open the Command Menu:
  - On Windows or Linux, press Control + Shift + P.
  - On macOS, press Command + Shift + P.
- Start typing JavaScript and select Disable JavaScript from the list, then press Enter.
- JavaScript will now be disabled for that specific tab as long as DevTools remains open.
- See that the home page loads all the information with JavaScript enabled.
- To re-enable JavaScript, open the Command Menu again and run the Enable JavaScript command, or simply close DevTools. 

## Issue / Card
https://github.com/canonical/canonical.com/issues/2118#issue-3698545694

Fixes #
https://warthogs.atlassian.net/browse/WD-31120

## Screenshots

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -162,7 +162,9 @@
       <div class="p-homepage-carousel">
         <div class="row">
           <div class="col-3 col-medium-6 p-homepage-carousel-tabs--wrapper">
-            <ul class="p-homepage-carousel-tabs u-hide--medium u-hide--small" role="tablist" aria-label="Homepage carousel tabs">
+            <ul class="p-homepage-carousel-tabs u-hide--medium u-hide--small"
+                role="tablist"
+                aria-label="Homepage carousel tabs">
               <li>
                 <a href="#start-ubuntu-tab"
                    class="p-homepage-carousel-tabs__item p-side-navigation__link"
@@ -278,7 +280,7 @@
                   <div class="progress-bar u-hide--large"></div>
                 </div>
                 <noscript>
-                    <h3 class="p-heading--4">Start with Ubuntu</h3>
+                  <h3 class="p-heading--4">Start with Ubuntu</h3>
                 </noscript>
                 <h3 class="u-hide--large p-heading--4 js-dependent">Start with Ubuntu</h3>
                 <p>
@@ -312,7 +314,7 @@
                   <div class="progress-bar u-hide--large"></div>
                 </div>
                 <noscript>
-                    <h3 class="p-heading--4">Take control of your infrastructure</h3>
+                  <h3 class="p-heading--4">Take control of your infrastructure</h3>
                 </noscript>
                 <h3 class="u-hide--large p-heading--4 js-dependent">Take control of your infrastructure</h3>
 
@@ -348,7 +350,7 @@
                   <div class="progress-bar u-hide--large"></div>
                 </div>
                 <noscript>
-                    <h3 class="p-heading--4">Build apps at speed and stay secure</h3>
+                  <h3 class="p-heading--4">Build apps at speed and stay secure</h3>
                 </noscript>
                 <h3 class="u-hide--large p-heading--4 js-dependent">Build apps at speed and stay secure</h3>
                 <p>
@@ -382,7 +384,7 @@
                   <div class="progress-bar u-hide--large"></div>
                 </div>
                 <noscript>
-                    <h3 class="p-heading--4">Get an optimized stack for data and AI</h3>
+                  <h3 class="p-heading--4">Get an optimized stack for data and AI</h3>
                 </noscript>
                 <h3 class="u-hide--large p-heading--4 js-dependent">Get an optimized stack for data and AI</h3>
 
@@ -420,7 +422,7 @@
 
                 </div>
                 <noscript>
-                    <h3 class="p-heading--4">Deploy smart things at the edge</h3>
+                  <h3 class="p-heading--4">Deploy smart things at the edge</h3>
                 </noscript>
                 <h3 class="u-hide--large p-heading--4 js-dependent">Deploy smart things at the edge</h3>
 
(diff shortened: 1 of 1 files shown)


eliman11:
Thanks @Onibenjo! I'll also tag @kim-isaac for review - we realized that the homepage breaks when JS is disabled, especially with the tabbed sections. The quick fix for this was to expose all the content on the page and having the tabs as anchor links instead. Would you mind giving this a look? I used [this](https://chromewebstore.google.com/detail/javascript-toggle-on-and/cdcgbgnfhhdmdkallfmlachogpghifgf?pli=1) extension to disable JS

I'm not sure what's feasible but if possible - 
- [ ] Could we hide these tabs as the tabs act as anchor links but isn't sticky, which means you'd have to scroll back up to use these tabs? Also the active state doesn't change here which is a bit confusing
Screenshot 2025-11-24 at 12 39 49

- [ ] Similarly could we hide these tabs as they don't function as anchor links and the content is already visible on the page without needing anchor links anyway?
Screenshot 2025-11-24 at 12 41 41
Screenshot 2025-11-24 at 12 39 58

kim-isaac:
Thanks @eliman11 for ping me here! @Onibenjo Also thanks for spotting the issue and addressing it. Together with Elaine’s comments, I’m providing the visual-related requests below.

- [ ] Would it be possible to make the [static suru](https://vanillaframework.io/docs/patterns/suru) is used as a fallback when JS is disabled?
- [ ] For the 'Innovate on your own terms' section's tabs, could we add a scroll offset(56px) for the fixed global navigation?
- [ ] Could we add regular padding between the subsections of “Accelerate innovation on any platform” and “News and insights from the source” like the image below?
image
image

- [ ] As Elaine mentioned, it seems the side tabs aren’t needed for “Accelerate innovation on any platform” and “News and insights from the source.” If it can be hidden in the JS-disabled version, could we place the added subsection heading in the tab area instead to keep the 25/75 layout? If this adds unnecessary complexity, please feel free to disregard the suggestion. Here is an example screen below:
image

- [ ] I noticed that “Latest stories” doesn’t appear in the JS-disabled version of the News page, so I wanted to confirm if this is expected.
- [ ] Since the subsection heading is duplicated in the default view as well, could you also apply `data-aria-hidden-on-js-load` to this one?
image

templates/index.html
```diff
@@ -453,438 +494,435 @@ <h2>Do more with trusted open source</h2>
...
+          </div>
+          <div class="col-9 col-medium-6">
+            <div class="p-section">
+              <div class="row">
+                <div class="col-9 col-medium-6">
+                  <h3 class="p-heading--4">
+                    Unlocking economies of scale
+                    <br class="u-hide--small u-hide--medium" />
+                    with open source infrastructure
+                  </h3>
```
Onibenjo: modified due to formatting

templates/index.html
```diff
@@ -453,438 +494,435 @@ <h2>Do more with trusted open source</h2>
...
+            {{ image(url="https://assets.ubuntu.com/v1/fca18349-5.png",
+                            alt="",
+                            width="2464",
+                            height="1027",
+                            hi_def=True,
+                            attrs={"class": "p-image--container__image"},
+                            loading="lazy") | safe
+            }}
+          </div>
+        </div>
```
Onibenjo, line 912: down to this line (modified due to formatting).

From R543 to R925.

templates/index.html
```diff
@@ -143,21 +161,24 @@ <h2>Innovate on your own terms</h2>
     <div class="p-section">
       <div class="p-homepage-carousel">
         <div class="row">
-          <div class="col-3 col-medium-6">
+          <div class="col-3 col-medium-6" style="position:sticky;height:fit-content;top:4rem">
```
mtruj013: These styles can probs be moved to the stylesheet

templates/index.html
```diff
@@ -143,21 +161,24 @@ <h2>Innovate on your own terms</h2>
...
-                        id="start-ubuntu"
-                        role="tab"
-                        data-index="0"
-                        aria-selected="true"
-                        aria-controls="start-ubuntu-tab">
-                  <div class="progress-bar--husk"></div>
-                  <div class="progress-bar"></div>
-                  Start with Ubuntu
-                </button>
+                <a href="#start-ubuntu-tab" class="p-side-navigation__link u-no-padding">
```
mtruj013: Nesting buttons within a tags is invalid html (docs [here](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element:~:text=The%20a%20element%20can%20be%20wrapped%20around%20entire%20paragraphs%2C%20lists%2C%20tables%2C%20and%20so%20forth%2C%20even%20entire%20sections%2C%20so%20long%20as%20there%20is%20no%20interactive%20content%20within%20(e.g.%2C%20buttons%20or%20other%20links).%20This%20example%20shows%20how%20this%20can%20be%20used%20to%20make%20an%20entire%20advertising%20block%20into%20a%20link%3A)), wondering if these need to be buttons at all or if they can be refactored to be only a elements as we typically do for tabs

templates/index.html
```diff
@@ -453,438 +494,435 @@ <h2>Do more with trusted open source</h2>
...
+                            width="150",
+                            height="57",
+                            hi_def=True,
+                            loading="lazy") | safe
+            }}
+          </div>
+          <div class="col-9 col-medium-6">
+            <div class="p-section">
+              <div class="row">
+                <div class="col-9 col-medium-6">
```
mtruj013: Maybe not in scope since this is on prod and not related to js-vs not, but the extra row and col divs on lines 564-565 aren't needed since this is already within a row and defined cols on line 562

templates/index.html
```diff
@@ -453,438 +494,435 @@ <h2>Do more with trusted open source</h2>
...
-                      <i class="p-icon--arrow-right"></i></a>
-                  </div>
+    <div tabindex="0"
+          role="tabpanel"
+          id="enhance-security-tab"
+          aria-labelledby="enhance-security"
+          aria-hidden="false">
+      <div class="p-section">
+        <div class="row">
+          <div class="col-3" style="padding-top: 0.45rem;">
```
mtruj013: This padding nudge is repeated a few times, can we move it to the stylesheet for easier maintainability?

mtruj013 (changes requested):
Thanks @Onibenjo! A couple of comments and optional suggestions:

templates/index.html
Onibenjo: I see this is in production as well.
I will make the adjustment

templates/index.html
Onibenjo: I tried removing them and the layout broke
image

L565 wraps the title.

This is the overview
image

Onibenjo: how it is 
image

templates/index.html
Onibenjo: Fixed now. Thanks for pointing out

templates/index.html
Onibenjo: Fixed

templates/index.html
mtruj013: (not a blocker for this pr, just general clean up so feel free to ignore) Should be this (col-9 is here is the original wrapper, not the nested one):
```
          <div class="col-9 col-medium-6">
            <div class="p-section">
              <h3 class="p-heading--4">
                Unlocking economies of scale
                <br class="u-hide--small u-hide--medium" />
                with open source infrastructure
              </h3>
              <div class="p-image--container u-hide--large">
                {{ image(url="https://assets.ubuntu.com/v1/1a25e055-1.png",
                                      alt="",
                                      width="2464",
                                      height="1027",
                                      hi_def=True,
                                      attrs={
                                      "class": "p-image--container__image",
                                      },
                                      loading="lazy") | safe
                }}
              </div>
              <div class="row">
                <div class="col-6 col-medium-6">
                  <p>
                    <i>“Canonical’s open source software sits at the heart of BT Group’s network cloud infrastructure, underpinning our 5G mobile core capability and providing the stability and flexibility required to support the increasing convergence of BT services with confidence.”</i>
                  </p>
                  <a class="p-cta-text u-hide--medium u-hide--small"
                      href="https://ubuntu.com/engage/5G-with-open-source-infrastructure">Learn how BT unlocked scalability<i class="p-icon--arrow-right"></i></a>
                </div>
                <div class="col-3 col-medium-3">
                  <p class="u-no-margin--bottom">
                    <b>Greg McCall</b>
                  </p>
                  <p class="u-text--muted">
                    Chief Networks Officer,
                    <br />
                    BT Group
                  </p>
                </div>
                <a class="p-cta-text u-hide--large"
                    href="https://ubuntu.com/engage/5G-with-open-source-infrastructure">Learn how BT unlocked scalability<i class="p-icon--arrow-right"></i></a>
              </div>
            </div>
          </div>
```

The heading is already wrapped in the original col-9 col-med-6 so it's set twice unnecessarily, the reason it breaks is because the nested row needs to be moved down for the quote layout (missed this bit when I left the original comment, my bad!). When we're already in a col-x we only need to define a new row when we need to break that down into even smaller columns. Eg here, we have a row with a col-3 for the links on the left, the rest is a col-9 ,so far so good no new row+cols definitions needed to serve design, then we get to the quote section which we need to break down further into a col-6 for the quote text and a col-3 for the author and it's here that we set a new row for those col definitions 

Image

Onibenjo: I see
Thanks for explaining this.
I have updated the code with your suggestion

PR #2090, opened 2025-11-26
WD-31128 Migrated resources page to /microk8s/resources

immortalcodes:
## Done

- Migrated page
- Fixed issues with side navigation

## QA

- [Demo](https://canonical-com-2090.demos.haus/microk8s/resources)
- Match the page with https://microk8s.io/resources

## Issue / Card

Fixes # https://warthogs.atlassian.net/browse/WD-31128

## Screenshots

### DIFF
--- a/templates/microk8s/resources/index.html
+++ b/templates/microk8s/resources/index.html
@@ -9,7 +9,7 @@ about Kubernetes from some of the industry's best.
 
 {% block meta_copydoc %}
 https://docs.google.com/document/d/1PB5-CamGmSjhbuyuGSariC9zUW-8krAJD0l8JyvNk20/edit#
-{% endblock%}
+{% endblock %}
 
 {% block body_class %}
 is-paper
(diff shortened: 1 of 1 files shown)


mattea-turic:
@immortalcodes, tbh I actually prefer it! 

I just wonder if we could perhaps fix it, as it doesn't really serve its function if it's moves with the rest of the content. Something like [this](https://www.figma.com/design/0DTIS3uNNUuDAdCKWrk8Uf/23.10-Microk8s.io--rebranding----Sites?node-id=2625-3438&t=bsf4B6RKRz5GQb9I-1) (added screenshot below too)? – but I'd like to get @eliman11's input and also add that if it complicates implementation a lot, that it's okay to leave it be!

Screenshot 2025-11-27 at 10 31 18 AM

immortalcodes:
> Thanks @immortalcodes! Agree with @mattea-turic if we can have the navigation sticky that would be ideal. Is there a reason we use this accordion rather than the "toggle side navigation" button as I thought that was the default responsive behaviour for that component?

We had picked that component from microk8s.io repo which had some custom properties that didn't go well in the current repo setup(we have multiple layers of navigation on top as opposed to microk8s) and hence I had to come up with a new approach.
Making it sticky seems a good idea, I will work on that.
Thanks a lot.

PR #2116, opened 2025-12-05
ci: Run deployment workflows sequentially to avoid race conditions

### LINKED ISSUE #2113: Deployment workflow potentially vulnerable to race conditions
The [deploy workflow](https://github.com/canonical/canonical.com/blob/022e09f9f125653edb1ed5a51ec38cf1c5c4bf18/.github/workflows/deploy.yaml) has no [concurrency group configuration](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency). This presents a race condition risk. If a PR lands while another is still deploying, we will have two deploy jobs running at the same time. If the deployment jobs pass in a different order than the PRs were merged, we may lose changes on the live site until a subsequent PR that includes both changes is deployed. 

### Example
PR A merges before PR B, and does not include PR B's changes. PR B  merges `main` in due to branch protection policies, then merges into main. There are now two deploy jobs running.

PR A's deploy job finishes **after** PR B's deploy job, but it doesn't include PR B's changes. So, PR B's changes are not live anymore. This could happen, for instance, if PR A's deployment hangs, fails and is manually retried, or by some processor priority fluke at the datacenter PR B's job completes first.

### Potential solution
[GitHub Actions concurrency groups](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency) are designed for this problem.

See Vanilla's deploy job: https://github.com/canonical/vanilla-framework/blob/main/.github%2Fworkflows%2Fdeploy.yml

petesfrench:
## Done

ci: Run deployment workflows sequentially to avoid race conditions

## QA

- The [workflow on vanilla-framework](https://github.com/canonical/vanilla-framework/blob/main/.github%2Fworkflows%2Fdeploy.yml#L11) was used as a reference. Compare their workflow to the additions here. 
- You can read about  GitHub action concurrency [here](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency)

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-31885
Fixes #2113

### DIFF
--- a/.github/workflows/a11y.yaml
+++ b/.github/workflows/a11y.yaml
@@ -1,0 +1,30 @@
+# This workflow setup only checks accessibility on the canonical.com homepage
+name: a11y check on canonical.com homepage
+
+on:
+  schedule:
+    - cron: "0 0 * * 1"  # every Monday at 00:00 UTC
+
+jobs:
+  check-a11y:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+      - name: Install dependencies
+        run: yarn
+
+      - name: Run a11y tests
+        run: |
+          yarn pa11y-ci
+
+      - name: Send message on failure
+        env:
+          BOT_URL: ${{ secrets.BOT_URL }}
+        if: failure()
+        run: curl -X POST -F "workflow=${GITHUB_WORKFLOW}" -F "repo_name=${GITHUB_REPOSITORY}" -F "action_id=${GITHUB_RUN_ID}" $BOT_URL?room=a11y-alerts
--- a/.pa11yci
+++ b/.pa11yci
@@ -1,0 +1,23 @@
+{
+  "defaults": {
+    "timeout": 120000,
+    "wait": 5000,
+    "chromeLaunchConfig": {
+      "args": [
+        "--no-sandbox"
+      ]
+    },
+    "ignore": [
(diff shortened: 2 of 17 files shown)


.github/workflows/deploy.yaml
```diff
@@ -123,6 +126,9 @@ jobs:
 
   deploy-production:
     runs-on: [self-hosted, self-hosted-linux-amd64-jammy-private-endpoint-medium]
+    concurrency: 
+      group: production-deployment
+      cancel-in-progress: false
     needs: [pack-charm, publish-image]
     steps:
       - name: Checkout Code
```
jmuzina: This reduces the race condition risk, but I think races might still be possible since you've applied the concurrency at the deploy job level rather than the workflow level. 

If PR A lands before PR B, but PR B manages to get all the way through rock packing to starting the deployment job before PR A does, then PR A will wait for PR B to finish deploying. Then, it will deploy the rock from PR A, erasing the changes from PR B. 

Maybe we need to move the concurrency group to the workflow level so that the entire deployment workflow is atomic.

petesfrench: You are right, I will move the concurrency group to the workflow level.

.github/workflows/deploy.yaml
```diff
@@ -123,6 +130,9 @@ jobs:
 
   deploy-production:
     runs-on: [self-hosted, self-hosted-linux-amd64-jammy-private-endpoint-medium]
+    concurrency: 
+      group: production-deployment
```
samhotep: @petesfrench do we still need these groups now that we have the one defined at the worflow level?

PR #2163, opened 2026-01-07
Update section in favor of linked logo pattern

mtruj013:
## Done

- Updated section to use linked logo pattern

## QA

- View the site locally in your web browser at: https://canonical-com-2163.demos.haus/careers/engineering (or any other careers page with that has the socials section)
- See that it has been updated to match suggested [design](https://warthogs.atlassian.net/browse/WD-20469?focusedCommentId=884391) (copy confirmed by @Sophie-32)

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-20469

## Screenshots

image

### DIFF
--- a/templates/contact-us.html
+++ b/templates/contact-us.html
@@ -91,19 +91,6 @@
         </div>
       </div>
     </div>
-    <div class="p-section--shallow">
-      <div class="row">
-        <hr class="p-rule" />
-        <div class="col-3 col-medium-2">
-          <h2 class="p-muted-heading">Legal notices</h2>
-        </div>
-        <div class="col-9 col-medium-4">
-          <p>
-            Please note that we cannot provide you with legal advice. Legal notices should be sent to <a href="mailto:legal@canonical.com">legal@canonical.com</a>. Questions about hiring, products or sales should not be sent to this address.
-          </p>
-        </div>
-      </div>
-    </div>
   </section>
   <section class="p-section">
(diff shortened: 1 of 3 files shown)


templates/careers/_base-department.html
```diff
@@ -90,35 +90,58 @@ <h2>{{ block_3_subheading }}</h2>
...
+  {% from "_macros/vf_linked-logo-section.jinja" import vf_linked_logo_section %}
+
+  {{ vf_linked_logo_section(
+    title_text="See for yourself the work we're doing",
+    layout="50/50",
+    links=[
+      {
+        "href": "https://github.com/canonical",
+        "text": "Follow Canonical on GitHub &rsaquo;",
+        "label": "Follow Canonical on GitHub",
```
muhammad-ali-pk: Can we omit the label, since it's identical to the paragraph below it? This will cause screen readers to read same text twice

mtruj013: Nice catch, you're right that the label isn't needed since the link text is already what we need (normally the two are different in this pattern). But just to clarify this would not cause screen readers to read both the label and the text as the aria-label takes precedence here and would become the accessible name. If that weren't so then every time this pattern is used both the text and label would be read out which isn't the case

muhammad-ali-pk: Ah, thanks for the clarification!

PR #2238, opened 2026-02-03
feat: attach utms cookie value to ubuntu.com redirects

britneywwc:
## Done

- Attach `utms` cookie value to ubuntu.com redirects

## QA

- Go to https://canonical-com-2238.demos.haus/?utm_content=test1&utm_medium=test2
- Check that utm contents are stored in `utms` cookie
- Scroll down to "Webinar & Events", click on any of the links that are redirecting to ubuntu.com
- Once redirected, check that the `utms` cookie values are appended in the URL

## Issue / Card

Fixes [WD-32640](https://warthogs.atlassian.net/browse/WD-32640)

## Screenshots

[WD-32640]: https://warthogs.atlassian.net/browse/WD-32640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -6,6 +6,7 @@ import datetime
 from geopy.geocoders import Nominatim
 from geopy.distance import geodesic
 from cachetools import TTLCache, cached
+import re
 
 
 def json_asset_query(file_name):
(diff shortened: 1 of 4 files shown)


mtruj013:
@britneywwc after accepting cookies I have to refresh twice for the cookie to be set, tested on both chrome and ff. Are you seeing the same?

webapp/views.py
```diff
@@ -16,3 +17,27 @@ def json_asset_query(file_name):
...
+        cookie_value = flask.request.cookies.get("utms")
+
+        if cookie_value:
+            data = response.get_data(as_text=True)
+            # Find all href attributes pointing to ubuntu.com
+            pattern = r'href=["\']([^"\']*ubuntu\.com[^"\']*)["\']'
+
+            def add_cookie_to_url(match):
+                url = match.group(1)
+                separator = "&" if "?" in url else "?"
```
mtruj013, line 319: Would you know if we can expect any key dupes? If so might be good to include a check so that the values get merged rather than potentially appending an already existing key

mtruj013 (approved):
LGTM, just a q/suggestion

webapp/views.py
britneywwc: Thanks, this is a good suggestion! The logic for handling duplicated keys from URL params should be handled in the `cookie-callback-with-policy.js` file. I created a separate ticket [here](https://warthogs.atlassian.net/browse/WD-33551) to address this

PR #2256, opened 2026-02-06
Fix a 11 y tests for the form generator template.

Skazitron:
## Done
- Fixed pa11y tests for canonical.com that were formerly ignored

## QA
- Go to [/#get-in-touch](https://canonical-com-2256.demos.haus/#get-in-touch)
- Everything functions normally
- Check that the legend tag visible to screenreaders doesn't interfere with the website.

## Issue / Card
https://warthogs.atlassian.net/browse/WD-31831

### DIFF
--- a/templates/lxd/index.html
+++ b/templates/lxd/index.html
@@ -7,7 +7,7 @@
 {% block meta_description %}
   Low-touch virtual infrastructure with LXD at any scale. Built on top of Linux containers,
   LXD offers a unified user experience in managing system containers, virtual machines and clusters. Easily configurable,
-  flexible and secure way to run your workloads.
+  flexible and trusted way to run your workloads.
 {% endblock %}
 
 {% block body_class %}
@@ -21,12 +21,12 @@
       <div class="row--25-75">
         <div class="col p-section">
           <div class="p-section--shallow">
-            <h1 class="u-no-margin--bottom">Lightweight open source virtualisation with LXD</h1>
-            <h2>The comprehensive platform for open source virtualisation with a cloud-like experience.</h2>
+            <h1 class="u-no-margin--bottom">Lightweight open source virtualization with LXD</h1>
+            <h2>The comprehensive platform for open source virtualization with a cloud-like experience.</h2>
           </div>
           <div class="p-section--shallow">
             <p>
-              LXD is an open source solution for managing virtual machines and system containers. It provides a secure and scalable environment with minimal overhead. Manage your workloads with ease and configure them to suit your use case via a user-friendly web interface.
+              LXD is an open source solution for managing virtual machines and system containers. It provides a security-maintained and scalable environment with minimal overhead. Manage your workloads with ease and configure them to suit your use case via a user-friendly web interface.
             </p>
           </div>
           <a class="p-button--positive" href="/lxd/install">Install now</a>
@@ -121,7 +121,7 @@
           <div class="col-3 col-medium-2">
             <h3 class="p-heading--5">Run any type of workload</h3>
             <p>
-              LXD provides both KVM-based VMs and system containers based on LXC – that can run a full Linux OS – in a single open source virtualisation platform. LXD has numerous built-in management features, including live migration, snapshots, resource restrictions, projects and profiles, and governs the interaction with various storage and networking options.
+              LXD provides both KVM-based VMs and system containers based on LXC – that can run a full Linux OS – in a single open source virtualization platform. LXD has numerous built-in management features, including live migration, snapshots, resource restrictions, projects and profiles, and governs the interaction with various storage and networking options.
 
             </p>
           </div>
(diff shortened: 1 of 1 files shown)


.pa11yci
```diff
@@ -7,17 +7,9 @@
...
-      "WCAG2AA.Principle4.Guideline4_1.4_1_2.H91.Textarea.Name",
-      "WCAG2AA.Principle4.Guideline4_1.4_1_2.H91.InputText.Name",
-      "WCAG2AA.Principle1.Guideline1_1.1_1_1.H65",
-      "WCAG2AA.Principle2.Guideline2_4.2_4_1.H64.1.NoTitle"
-    ],
     "hideElements": "iframe[style*='display: none'], [style*='visibility: hidden;']"
   },
   "urls": [
-    "https://canonical.com/"
+    "http://0.0.0.0:8002"
```
Skazitron: this is set to local right now for testing purposes, but will be switched to canonical.com

templates/shared/forms/form-fields.html
```diff
@@ -22,10 +22,16 @@
...
+            <span class="u-off-screen">{{ fieldset.title }}</span>
+          </legend>
           <div class="row--50-50 {% if fieldset.noCommentsFromLead != True %}js-formfield{% endif %}">
             <div class="col">
-              <legend class="p-heading--4 js-formfield-title {% if fieldset.isRequired %}is-required{% endif %}"
-                      id="{{ fieldset_id }}-legend">{{ fieldset.title }}</legend>
+              <p class="p-heading--4 js-formfield-title {% if fieldset.isRequired %}is-required{% endif %}"
+                  aria-hidden="true">
+                  {{ fieldset.title }}
+              </p>
```
britneywwc: This approach works as pa11y requires `legend` to be a direct child of `fieldset`. The issue is that by using `p` tag for the fieldset title, it removes stylings that we have for the `is-required` class. 

We could update [vanilla-framework](https://github.com/canonical/vanilla-framework/) to include `*` for `is-required` classnames or find a different approach to keep the legend stylings while also preserving the a11y requirements. The only issue I could think of for the former approach is that `p` might not have a robust use for using this custom `is-required` classname in other cases.

britneywwc (commented):
Thanks for this piece of work! Left a comment/question regarding the legend stylings

static/sass/styles.scss
```diff
@@ -743,6 +743,11 @@ $p-small-lh-diff: map-get($line-heights, default-text) -
   }
 }
 
+.p-is-required::before {
```
britneywwc:
```suggestion
.p-heading--4.legend-is-required::after {
```
Since this custom style is dependant on the `is-required` class, it would be better if we attached the custom classname to the required class rather than `p`

templates/shared/forms/form-fields.html
```diff
@@ -22,10 +22,13 @@
...
         <fieldset class="p-fieldset-section{% if fieldset.inputType == 'checkbox-visibility' %} js-toggle-checkbox-visibility{% elif fieldset.isRequired and fieldset.inputType == 'checkbox' %} js-required-checkbox{% elif fieldset.inputType == 'radio' %} js-remove-radio-names{% endif %} {% if fieldset.inputType == 'checkbox' or fieldset.inputType == 'checkbox-visibility' %}js-remove-checkbox-names{% endif %}"
                   aria-labelledby="{{ fieldset_id }}-legend">
+            <legend class="u-off-screen">
+                {{ fieldset.title }}{% if fieldset.isRequired %} (required) {% endif %}
+            </legend>
           <div class="row--50-50 {% if fieldset.noCommentsFromLead != True %}js-formfield{% endif %}">
             <div class="col">
-              <legend class="p-heading--4 js-formfield-title {% if fieldset.isRequired %}is-required{% endif %}"
-                      id="{{ fieldset_id }}-legend">{{ fieldset.title }}</legend>
+              <p class="p-heading--4 js-formfield-title {% if fieldset.isRequired %}p-is-required{% endif %}"
```
britneywwc:
```suggestion
              <p class="p-heading--4 js-formfield-title {% if fieldset.isRequired %}legend-is-required{% endif %}"
```

static/sass/styles.scss
Skazitron: @britneywwc  got it. will have to use the ::before pseduo element though since it's closer to the vanilla framework representation.

static/sass/styles.scss
```diff
@@ -743,7 +743,7 @@ $p-small-lh-diff: map-get($line-heights, default-text) -
   }
 }
 
-.p-is-required::before {
+.p-heading--4.legend-is-required::before {
```
britneywwc, line 746: Missed this in my previous review sorry, but there should be a space between `*` and the title

Skazitron: @britneywwc done!

PR #2313, opened 2026-02-25
feat: handle duplicated utm cookie values

britneywwc:
## Done

- Handle duplicated `utm_` cookie values by concatenation
- Decode URI cookie values before appending them to ubuntu.com

## QA
- Go to https://canonical-com-2313.demos.haus/?utm_content=test1&utm_medium=test2&utm_medium=test3
- Check that `utms` cookie is set and concatenates duplicated `utm_medium` values as such: 
`utm_content:test1&utm_medium:test2`
- Inspect page and do a search for "https://ubuntu.com"
- See that the utm values are concatenated on the link

## Issue / Card

Fixes [WD-33551](https://warthogs.atlassian.net/browse/WD-33551)

## Screenshots

[if relevant, include a screenshot]

[WD-33551]: https://warthogs.atlassian.net/browse/WD-33551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2316, opened 2026-02-26
feat: append utm cookies before hash tags

britneywwc:
## Done

- Add a check to append hashes/jump tags after appending utms
- Addresses [this PR comment](https://github.com/canonical/canonical.com/pull/2313#issuecomment-3959090120)

## QA

- Go to https://canonical-com-2316.demos.haus/?utm_content=test1&utm_medium=test2&utm_content=test2
- Click on "Discover our solutions"
- See that it redirects and appends utm values as expected

## Issue / Card

Fixes #

## Screenshots

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -3,7 +3,7 @@ import flask
 import requests
 import math
 import datetime
-from urllib.parse import unquote
+from urllib.parse import urlparse, urlunparse, unquote
 from geopy.geocoders import Nominatim
 from geopy.distance import geodesic
 from cachetools import TTLCache, cached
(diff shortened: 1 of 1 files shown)


webapp/views.py
```diff
@@ -311,16 +310,24 @@ def append_utms_cookie_to_ubuntu_links(response):
...
             pattern = r'href=["\']([^"\']*ubuntu\.com[^"\']*)["\']'
 
             def add_cookie_to_url(match):
                 url = match.group(1)
-                separator = "&" if "?" in url else "?"
-                new_url = f"{url}{separator}{cookie_value}"
+
+                # Check if URL has a hash and insert UTM before it
+                if "#" in url:
+                    base_url, hash_part = url.split("#", 1)
```
Skazitron: Can you please use the urllib.parse library for this?

Skazitron:
```def add_cookie_to_url(match):
    url = match.group(1)
    
    # 1. Break the URL into components
    # (scheme, netloc, path, params, query, fragment)
    u = urlparse(url)
    
    # 2. Convert the query string into a list of key-value pairs
    query_params = parse_qsl(u.query)
    
    # 3. Add your cookie/UTM logic
    # Assuming cookie_value is something like "utm_source=my_cookie"
    # If it's just a value, use: query_params.append(('cookie', cookie_value))
    # For this example, let's assume cookie_value is 'user=123'
    key, _, value = cookie_value.partition('=')
    query_params.append((key, value))
    
    # 4. Reconstruct the query string and the full URL
    new_query = urlencode(query_params)
    new_url = urlunparse((
        u.scheme, 
        u.netloc, 
        u.path, 
        u.params, 
        new_query, 
        u.fragment
    ))
    
    return f'href="{new_url}"'```

Comment (Skazitron) [reply webapp/views.py]: I think doing manual string parsing like that might be prone to edge cases

Comment (britneywwc) [reply webapp/views.py]: Thanks, this is a good suggestion!

File: templates/index.html [hunk 2]
```diff
@@ -78,7 +78,7 @@ <h1 class="p-heading--display" id="suru-motion-anchor">
           <p>Simple, cost-effective, supported &ndash; expand the Ubuntu philosophy to every layer of your enterprise stack.</p>
         </div>
         <div style="z-index: 3; position: relative">
-          <a class="p-button--positive" href="/solutions">Discover our solutions</a>
+          <a class="p-button--positive" href="https://ubuntu.com/download/core?test=1#release-notes">Discover our solutions</a>
```
britneywwc: Hardcoded for QA purposes. Revert before merging

PR #2354, opened 2026-03-12
chore: filtering sentry error to exclude upstream connection failures

Onibenjo:
- sentry_before_send was only filtering werkzeug.exceptions.HTTPException (Flask's own HTTP errors). When the app makes upstream HTTP requests using the requests library and receives a 4xx response, requests raises requests.exceptions.HTTPError — a completely different exception class with no relationship to Werkzeug's hierarchy. These errors were slipping through the filter and appearing in Sentry noise (e.g. 401 Client Error from upstream APIs).
- the WP API returns repeated Connection Error and it's being unhandled — resulting in noisy unhandled exceptions in Sentry.

## Done
- Extend sentry_before_send to sample out 95% of blog/WordPress API retry errors matching /wp-json/wp/v2.
- Added a second isinstance check for requests.exceptions.HTTPError in sentry_before_send. Since this exception stores the response on .response.status_code (not .code), it requires its own check

## QA

- Check out this [demo](https://canonical-com-2354.demos.haus/)
- Check that it loads up

## Issue / Card

Fixes #

## Screenshots

### DIFF
--- a/tests/test_sentry_filtering.py
+++ b/tests/test_sentry_filtering.py
@@ -161,7 +161,7 @@ class TestSentryFiltering(unittest.TestCase):
 
     def test_allow_non_blog_connection_error(self):
         """
-        Test that connection errors unrelated to the blog API 
+        Test that connection errors unrelated to the blog API
         always pass through.
         """
         exc = RequestsConnectionError("upstream connection failed")


webapp/app.py
```diff
@@ -287,6 +289,20 @@ def sentry_before_send(event, hint):
...
+            response = getattr(exc_value, "response", None)
+            if response is not None and 400 <= response.status_code < 500:
+                return None
+        # Sample 5% of transient upstream connection failures
+        # (e.g. WordPress API being unavailable)
+        if isinstance(exc_value, (MaxRetryError, RetryError, ConnectionError)):
+            error_msg = str(exc_value)
+
+            # Sample blog/WordPress API retry errors
+            if "/wp-json/wp/v2" in error_msg:
```
petesfrench, line 303: This is a magic string. Could we define it at a higher level? Or is it defined already somewhere else?

Onibenjo: It's not defined elsewhere.

And it's only used on this line so there might be no need for a variable

PR #2404, opened 2026-03-26
Fix sitemap missing file when generation fails

Onibenjo:
## Done
- Add missing test coverage for build_sitemap_tree
- Removed https://canonical.com/juju/docs/sitemap.xml from robots.txt as it doesn't exist
- Return an error if sitemap generation fails

## QA

- Open the [DEMO](https://canonical-com-2404.demos.haus/)
- Check that sitemap loads without issue: https://canonical-com-2404.demos.haus/sitemap_tree.xml
- Alternatively, check out this feature branch

## Issue / Card

Fixes #

## Screenshots

### DIFF
--- a/tests/test_sitemap.py
+++ b/tests/test_sitemap.py
@@ -130,7 +130,10 @@ class TestSitemap(unittest.TestCase):
         result = create_sitemap("/invalid/path/sitemap_tree.xml")
         self.assertEqual(result, ({"error:", "Sitemap is empty"}, 400))
 
-    @patch("webapp.app.directory_parser.generate_sitemap", return_value="<urlset/>")
+    @patch(
+        "webapp.app.directory_parser.generate_sitemap",
+        return_value="<urlset/>",
+    )
     @patch("builtins.open", unittest.mock.mock_open(read_data="<urlset/>"))
     @patch(
         "webapp.app.os.path.exists",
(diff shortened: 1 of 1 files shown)


webapp/app.py
```diff
@@ -1687,6 +1687,10 @@ def serve_sitemap():
         if not os.path.exists(sitemap_path):
             xml_sitemap = create_sitemap(sitemap_path)
 
+        # If still missing (generation failed), return an error
+        if not os.path.exists(sitemap_path):
+            return {"error": "Sitemap not available"}, 503
```
samhotep, line 1692: Looks good, though do you think a `404` is more appropriate response?

Onibenjo: The sitemap should auto generate and if it doesn't, it means something was wrong. and we should investigate I think.
And it shows a server error and not a "not found"

Onibenjo: in this case a 404 is usually the wrong signal.

- GET /sitemap_tree.xml is a real, defined endpoint. When it can’t produce the sitemap because generation failed (or the filesystem write didn’t happen), that’s a server-side failure, not “resource not found”.
- A 404 tells clients (and search engines) the sitemap doesn’t exist (potentially permanent), which can cause the sitemap URL to be dropped.
- A 503 is more appropriate, since we genuinely expect it to be temporary and retryable

samhotep: I see, okay then

PR #2447, opened 2026-04-12
fix: udpate lxd docs paths

minaelee:
Updates LXD documentation paths to use the docs version corresponding to the current LTS release, and to remove the no-longer-used /en/ part of the path.
Also fixes a link in https://canonical.com/lxd/install - the link for "Learn more about LXD and security" was pointing to the docs on containers and VMs instead of the security docs.


elijahgreenstein:
This PR can be closed, as these changes have been requested through other channels.

PR #2475, opened 2026-04-23
fix: add knowledge categories and remove duplicated anchors

britneywwc:
## Done

- Add two categories to knowledge hub
- Remove duplicated anchor heading

## QA

- Go to https://canonical-com-2475.demos.haus/knowledge/security-and-compliance/security-vs-compliance or any other knowledge articles
- Click on a link under the in-page-navigation
- See that the anchor link does not have "-2" appended to it
- For the category fix, see that the two parent categories match the entries on the [tag taxonomy sheet](https://docs.google.com/spreadsheets/d/1999CK3b1DbZuwj0CRfVX3pdYSdugeG1ALw0Wki8Vmgg/edit?gid=0#gid=0)

## Issue / Card

Fixes [WD-36139](https://warthogs.atlassian.net/browse/WD-36139)

## Screenshots

[WD-36139]: https://warthogs.atlassian.net/browse/WD-36139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/static/js/in-page-navigation.js
+++ b/static/js/in-page-navigation.js
@@ -311,21 +311,11 @@ function generateSelectors(navRoot) {
  * @returns {string} The heading ID
  */
 function generateHeadingId(heading) {
-  if (heading.id && !document.getElementById(heading.id)) {
+  if (heading.id) {
     return heading.id;
   }
 
-  let baseId = slugify(heading.textContent);
-  let id = baseId;
-
-  // Handle duplicate IDs
-  let counter = 1;
-  while (document.getElementById(id)) {
-    appendix = counter == 1 ? '' : `-${counter}`;
-    id = baseId + appendix;
-    counter++;
-  }
-
+  const id = slugify(heading.textContent);
   heading.id = id;
   return id;
 }
--- a/templates/knowledge/_base_knowledge_category.html
+++ b/templates/knowledge/_base_knowledge_category.html
@@ -47,7 +47,9 @@
         </div>
       {% else %}
         <div class="grid-row">
-          <p>No articles available in this category.</p>
+          <div class="grid-col-start-large-3 grid-col-6">
+            <p>No articles available in this category.</p>
+          </div>
         </div>
       {% endif %}
     </section>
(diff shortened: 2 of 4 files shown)


britneywwc:
@eliman11 Can we assume that the headings will never be of duplicated value? I removed the logic to append "-2" etc that was supposed to catch this edge case.

eliman11:
> @eliman11 Can we assume that the headings will never be of duplicated value? I removed the logic to append "-2" etc that was supposed to catch this edge case.

Thanks Britney! Yeah I think it's fair to assume we won't have duplicated headings.

Not a huge deal as these pages won't be indexed, but could we align the "No articles available.." text to the heading?
Screenshot 2026-04-23 at 09 34 59

britneywwc:
@eliman11 Thanks! Rebased the fix, it should align now

static/js/in-page-navigation.js
```diff
@@ -311,21 +311,11 @@ function generateSelectors(navRoot) {
...
-  if (heading.id && !document.getElementById(heading.id)) {
+  if (heading.id) {
     return heading.id;
   }
 
-  let baseId = slugify(heading.textContent);
-  let id = baseId;
-
-  // Handle duplicate IDs
-  let counter = 1;
```
Skazitron, line 322: Is deleting the logic for handling duplicate ID's safe? Can you please look up why this line was added from the commit logs and ensure it's okay to delete it?

eliman11:
@britneywwc sorry should've clarified - could we have the paragraph start at the 25% mark so it's left-aligned with the heading?

static/js/in-page-navigation.js
britneywwc: Yes, I checked with @eliman11 in the previous comment. It is safe for us to assume that there will be no duplicated headings for the knowledge articles. This was brought over from the [Vanilla sample script](https://github.com/canonical/vanilla-framework/blob/main/templates/docs/examples/patterns/in-page-navigation/_in-page-navigation.js) that handles duplicated headings.

Skazitron (approved):

britneywwc:
@eliman11 Addressed your design comment, can you have another look please? Will merge once I get the UX+1

Skazitron (approved):

PR #2506, opened 2026-05-04
Apply rebrand

mtruj013:
## Done

- Apply rebrand as per [figma](https://www.figma.com/design/rEDsQqLe8SfnvoxCcXJYLr/canonical.com-case-study---Sites?node-id=1881-3007&t=BZ4gpPzUDBEyI8Ge-0)
- Note: Differences to be expected as target dropdown multi-select component is not yet ready

## QA

- Open the [DEMO](https://canonical-com-2506.demos.haus/case-study)
- Check against design

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-20792

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -1149,3 +1149,15 @@ body:not(.js-enabled) {
   margin-left: 0.5rem;
   margin-right: 0.5rem;
 }
+
+	.p-case-study__filters {
+		display: flex;
+		gap: $spv--large;
+	}
+
+	.p-case-study__filter-select {
+		background-color: $color-paper;
+		border: none;
+		max-width: min-content;
+    text-overflow: ellipsis;
+	}
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -1150,14 +1150,54 @@ body:not(.js-enabled) {
   margin-right: 0.5rem;
 }
 
-	.p-case-study__filters {
-		display: flex;
-		gap: $spv--large;
-	}
+.p-case-study__filters {
+  display: flex;
+  gap: $spv--large;
+  @media screen and (width < $breakpoint-small) {
+    flex-direction: column;
+    padding-bottom: $spv--x-large;
+    
+    .p-case-study__filter-heading {
+      margin-bottom: 0;
+    }
+  }
+}
 
-	.p-case-study__filter-select {
-		background-color: $color-paper;
-		border: none;
-		max-width: min-content;
-    text-overflow: ellipsis;
-	}
+.p-case-study__filter-select {
+  background-color: $color-paper;
+  border: none;
+  max-width: min-content;
+  text-overflow: ellipsis;
+
+  @media screen and (width < $breakpoint-small) {
+    padding-left: 0;
+  }
+}
+
+.p-pagination__case-study {
+  @extend .p-pagination;
+  @media screen and ($breakpoint-large > width) {
+    justify-content: center;
+  }
+}
+
+.p-case-study__cards {
+  @media screen and (width < $breakpoint-small) {
+    .p-content-card__image-wrapper {
+      display: none;
+    }
+
+    .p-content-card__hover-content {
+      display: none;
+    }
+
+    .p-content-card {
+      height: auto;
+      min-height: unset;
+    }
+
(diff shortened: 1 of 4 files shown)


mtruj013:
Leaving placeholder image only as the actual images coming from case studies are all over the place 
Screenshot 2026-05-04 203922

petesfrench:
[ ] The first item's title is leading with a `Null`
image

[ ] Is it possible to reduce the spacing between 'Tag' and the chevron?
image

[ ] Can each individual tag have it's own chip

[ ] Only using the default canonical image looks a little off imo. WDYT @eliman11?

[ ] Can we center the 'no case studies found' message
image

mtruj013:
Thanks @petesfrench , aligned the "no case studies" found and cleaned up the template. For the rest:

> The first item's title is leading with a Null

 This isn't coming from this redesign, it's on the live version as well so likely something that needs to be fixed at the parser or individual discouse post level
Screenshot 2026-05-05 100020

> Is it possible to reduce the spacing between 'Tag' and the chevron?
 
Not with the native browser select, no. We agreed that we would be going with that for this version, comment [here](https://www.figma.com/design/rEDsQqLe8SfnvoxCcXJYLr?node-id=1881-2993#1744763029)

> Can each individual tag have it's own chip

Not with this pattern, the current macro renders as follows:
`<span class="p-chip--information is-readonly">{{ footer.content_type }}</span>`

> Only using the default canonical image looks a little off imo. WDYT @eliman11?

I agree, it isn't ideal, but the alternative didn't look much better tbh, relevant comment [here](https://github.com/canonical/canonical.com/pull/2506#issuecomment-4373562499). Happy to go with which ever one y'all this is best

eliman11:
Tysm @mtruj013! Re Pete's comments - 
> [ ] Is it possible to reduce the spacing between 'Tag' and the chevron?

Looks like this is because we're using the "select" component rather than the [base button](https://vanillaframework.io/docs/patterns/buttons#base) with the icon which hugs the text width rather than stay as a set width as the select component does. If this is an easy switch that would be great, but if not happy to refresh it in the future with the new dropdown styling. 

> [ ] Can each individual tag have it's own chip

Agree it would be nice but as we'll be refreshing the tags soon anyway as part of the engage page rebranding epic, we can leave this out for now.

Other than that could you please - 
- [x] Remove the rule above pagination
- [x] Wrap the section with the pagination in deep padding
- [x] On tablet and mobile could we have pagination center aligned on the page
- [x] On small screens could we have the "Filter by" "Language" and "Tag" stack vertically to avoid the wrapping in "Filter by"
Screenshot 2026-05-05 at 11 26 48

- [x] On medium screens there's this forced line break that makes the heading read awkwardly. Could we have it wrap as it would naturally
Screenshot 2026-05-05 at 11 27 52

- [x] Hide image thumbnails on small screens
- [ ] Is it possible to ensure we can navigate to the filters using keyboard navigation? Currently it's not possible to open the list and make a selection via keyboard

And lastly on thumbnails - is there a reason we aren't able to display the the images for these posts? It seems like there is a meta_image specified on the Discourse posts so I'm unsure why they aren't rendering correctly. Example: Airlock digital Portuguese case study ([Discourse](https://discourse.ubuntu.com/t/pt-case-study-airlock-digital-boosts-operational-performance-on-azure-by-30-with-an-os-and-support-model-that-just-works/56026), [image](https://assets.ubuntu.com/v1/a9601d48-PT%20-%20AIRLOCK%20Case%20Study.png)) or Grundium Portuguese case study ([Discourse](https://discourse.ubuntu.com/t/pt-case-study-grundium-secures-critical-systems-and-improves-health-outcomes-with-ubuntu-pro-for-devices/64879), [image](https://assets.ubuntu.com/v1/fc5e1f9b-PT%20-%20Grundium%20Case%20Study%20(1).png)). Thank you so much!!
Screenshot 2026-05-05 at 11 33 27
Screenshot 2026-05-05 at 11 31 49

eliman11:
Ah! Edit: Realized the images were a bit funky because you're pulling from the "image" field from Discourse as opposed to the "meta_image" field. I think if you make this change we'll have much cleaner thumbnails!

petesfrench (approved):

PR #2540, opened 2026-05-21
fix: add CSP nonce setup and apply nonces to inline scripts

Skazitron:
## Done
  - Added per-request `csp_nonce` generation in `webapp/handlers.py` (`before_request` + `context_processor`) and inject it into `script-src` / `script-src-elem` via a new
  `NONCED_DIRECTIVES` constant.
  - Added `'strict-dynamic'` to `script-src-elem` and removed `'unsafe-inline'` / `'unsafe-hashes'` from `script-src` and `script-src-elem`.
  - Added `nonce="{{ csp_nonce }}"` to every inline and external `<script>` tag across ~95 templates so they continue to execute under the tighter policy.
  - Refactored two inline scripts that could not stay as global function declarations under CSP:
    - `templates/careers/application/index.html` — `showInput()` rewritten as an IIFE wired up with `addEventListener('change', …)`; paired with removal of
  `onchange="showInput()"` in `_withdrawal-form.html`.
    - `templates/juju/docs/search.html` — `window.loadMoreResults = function() {…}` rewritten as `loadMoreButton.addEventListener('click', …)`; paired with removal of
  `onclick="loadMoreResults()"` on the button.
  
  ## QA
  - DevTools console should show **no CSP violations** on a sampling of pages: homepage, `/careers`, `/careers/<role>`, `/careers/application/<id>` (if accessible),
  `/juju/docs`, `/juju/docs/search?q=foo`, `/maas/features`, `/blog`, any legal page.
  - Reload twice and confirm the `nonce-…` value in the `Content-Security-Policy` response header changes per request, and that the `nonce=` attribute on inline `<script>`
  tags matches the header.
  - On `/careers/application/<id>` (or its withdrawal form): selecting "Other" as the withdrawal reason still shows the free-text input and marks it required; selecting any
  other option hides it again.
  - On `/juju/docs/search?q=<query>`: the "Load more results" button still reveals the next 10 results and switches to "Loading…" briefly.
  - Check Network → Headers: `Content-Security-Policy` no longer contains `'unsafe-inline'` (in `script-src` / `script-src-elem`) or `'unsafe-hashes'`, and contains
  `'strict-dynamic'` in `script-src-elem`.

### DIFF
--- a/templates/academy/exam-content/index.html
+++ b/templates/academy/exam-content/index.html
@@ -87,7 +87,7 @@
       </div>
     </div>
   </section>
-  <script>
+  <script nonce="{{ csp_nonce }}">
     (function() {
       var keys = {
         left: 'ArrowLeft',
--- a/templates/academy/exam-guide.html
+++ b/templates/academy/exam-guide.html
@@ -1026,7 +1026,7 @@ body_class %}
   </div>
 </section>
 
-<script>
+<script nonce="{{ csp_nonce }}">
   /**
     Toggles the necessary aria- attributes' values on the accordion panels
     and handles to show or hide them.
--- a/templates/academy/self-study.html
+++ b/templates/academy/self-study.html
@@ -488,7 +488,7 @@
       </div>
     </div>
   </section>
-  <script>
+  <script nonce="{{ csp_nonce }}">
     const setupTabs = () => {
       var keys = {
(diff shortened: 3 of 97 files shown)


Ataku-B (approved):
Thank you for the effort! LGTM! 👍

PR #2542, opened 2026-05-21
WD-36600: allow regional google.* domains in connect-src for GTM

Skazitron:
## Done
  - Added ~40 country-specific `www.google.*` hostnames (EU, Americas, APAC, MEA) to `connect-src` in `webapp/handlers.py` so the GTM `ads/ga-audiences` pixel can fire for
  visitors in those regions.
  
## QA
  - From a few geographies (or via VPN / Chrome `--lang` + cookie spoofing): load the homepage with the cookie banner accepted, then check DevTools → Network for the
  `ads/ga-audiences` request to `www.google.<tld>` and confirm it is not blocked by CSP.
  - Console should not show `Refused to connect to 'https://www.google.<tld>/…' because it violates the following Content Security Policy directive: "connect-src …"` for any
  of the listed domains.
  - Diff `Content-Security-Policy` response header against `main` — only `connect-src` should grow; no other directive changes.


Ataku-B (commented):
How do we decide which regional Google domains to use here? Google has a [supported domains list](https://www.google.com/supported_domains), could we use that as the source of regional Google domains?

Skazitron:
@Ataku-B Nice! I didn't know that google provided a list like that. I suppose we can use that instead of what Gemini spat out.

Skazitron:
@Ataku-B updated handlers.py to fetch google domains dynamically.

Ataku-B (approved):
👍 Approved! Thank you!

PR #2551, opened 2026-05-23
Change key value from 'event' to 'roadshow'

petesfrench:
## Done

- update roadshow event type query param

## QA

- Open the [DEMO](https://canonical-com-2551.demos.haus/events/canonical-days)
- Check you can see the roadshow events. Example:
image


Ataku-B (approved):
👍 LGTM!

PR #2597, opened 2026-06-09
fix: update inline style blocks and add nonces

Skazitron:
## Done
 - Unsafe-inline removed for styles and added a nonce mechanism
 - Migrated inline styles to dedicated css files.

## QA
- Open the demo or clone this branch and run it locally.
- Visit `Files changed`
- Go to the subroutes that are related to those files.
- Ensure there are no visual bugs related to changed CSS.
- Open the browser console and ensure that there are no errors related to the changed CSS.

## Issue / Card
https://warthogs.atlassian.net/browse/WD-36639


static/sass/_utilities_csp.scss
Ataku-B, line 1: I am not sure if naming utility classes by their content verbatim like this is the best way to do this. This approach has some issues such as:
- URL's in class names can cause headaches. If the URL of the asset changes, after changing the URL attribute in the class, we would have to rename the class everywhere it's used to keep consistency.
- There are floating point numbers to the 10th decimal used as class names such as `u-aspect-ratio-5-1775700934579-width-100`
- There are misleading class names such as `.u-margin-bottom-5rem` which actually applies a .5 rem margin not a 5rem margin, or `.u-margin-left-1rem` applying a -1 rem margin but the negative is lost in the naming

Overall these utility classes will be difficult to maintain and reuse. I think we should have them named based on usage rather than content, such as `.hero-background` or `.animated-header`. Also, again for the sake of re-usability, instead of having one class having 6 highly specific attributes for a single element, we can break down those into smaller classes and apply those multiple classes in the HTML.

Skazitron: Addressed your points:
- URLs are now only in the SCSS declaration.
- class names are less vague and misleading.
- Multi-property classes have been broken down into atomic single property classes.

Ataku-B (approved):
LGTM 👍

Ataku-B (approved):
👍 LGTM!

PR #2651, opened 2026-06-19
ci: pin copilot model

britneywwc:
## Done

- Pin copilot model to `gpt-5` for: strong instruction following, code-edit reliability with no additional thinking required
- The `.lock.yml` file is created by compiling the `md` file using `gh aw compile .github/workflows/copilot-assignment.md`

## QA

- No QA steps as this workflow can only be triggered once merged to main

## Issue / Card

Fixes #

## Screenshots


.github/workflows/copilot-assignment.md
britneywwc, line 1: Changes made to this file only

.github/workflows/copilot-assignment.md
```diff
@@ -20,6 +22,7 @@ safe-outputs:
   assign-to-agent:
     name: copilot
     target: triggering
```
immortalcodes, line 26: There two are already default values but again no-harm in mentioning them.
Also can we link this doc reference in comments, might be useful in the future.
https://github.github.com/gh-aw/reference/frontmatter-full/#complete-frontmatter-reference

Also a non-blocking suggestion, shall be use a variable name "model" instead of hard-coding gpt-5?

immortalcodes (approved):
LGTM! With a minor comment

.github/workflows/copilot-assignment.md
britneywwc: Thanks! Added the docs

I kept the model hardcoded on the file for now as it is specific for this workflow. If there is a need in the future for other agentic worfklow files to do the same then we can revisit it.

PR #2660, opened 2026-06-19
feat: audit and harden CSP headers, add report-only mode and violation reporting

immortalcodes:
## Done

Audited the site's Content-Security-Policy (`webapp/handlers.py`) and applied a safe, testable hardening pass:

- Added previously-missing defense-in-depth directives that don't change current behaviour
- Added `unpkg.com` to `connect-src` , it's the fallback CDN the homepage Lottie animations fetch from when jsdelivr fails, but was missing from the allowlist, so the fallback was silently broken.
- Added a `/csp-report` endpoint and `report-uri` directive
- Added a second `Content-Security-Policy-Report-Only` header (`CSP_REPORT_ONLY`) that additionally drops `'unsafe-eval'` and several vendor domains with zero remaining references anywhere in the codebase. This **does not block anything** , it's a no-risk way to watch real traffic via the new `/csp-report` endpoint/Sentry for a bake-in period before folding those removals into the enforced `CSP` dict for real.

Known gaps flagged but intentionally not changed in this PR:
- `frame-ancestors` / `X-Frame-Options` — no clickjacking protection currently exists.
- `img-src: *` wildcard — already documented in code as intentional, for Google Ads pixel images.

## QA

- Run the site locally.
- Open browser DevTools → Network/Console and confirm:
  - The response has both a `Content-Security-Policy` and a `Content-Security-Policy-Report-Only` header.
  - No new console errors/blocked-resource warnings appear on the homepage, careers page (web worker), and a Marketo form page (e.g. partner/contact-us forms) compared to `main`.
- Trigger a violation report end-to-end: open DevTools console and run
  `fetch("https://example.com")` on any page, then check the app logs for a `CSP violation report: ...` 

## Note:
- Tested out on sentry.
- Over the following days/weeks in production, watch Sentry/app logs for violations with `disposition: "report"` matching the six removed entries above (`crazyegg`, `zi-scripts`, `zoominfo`, `snap.licdn.com`, `buttons.github.io`, `tfaforms`, `'unsafe-eval'`) before promoting those removals into the enforced policy.

## Issue / Card

Ref: WD-36642

### DIFF
--- a/webapp/handlers.py
+++ b/webapp/handlers.py
@@ -1,8 +1,10 @@
 import logging
+import time
 
 import requests
 import secrets
 import sentry_sdk
+from urllib.parse import urlparse
 
 import flask
 from canonicalwebteam.flask_base.env import get_flask_env
@@ -222,6 +224,7 @@ CSP = {
         "'self'",
         "https://pages.ubuntu.com",
         "https://ubuntu.com",
+        "https://www.tfaforms.com",
     ],
     "object-src": ["'none'"],
     "base-uri": ["'self'"],
@@ -241,7 +244,6 @@ _CSP_REPORT_ONLY_REMOVALS = {
         "js.zi-scripts.com",
         "snap.licdn.com",
         "buttons.github.io",
-        "www.tfaforms.com",
     ],
     "connect-src": [
         "*.crazyegg.com",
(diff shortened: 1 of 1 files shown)


immortalcodes:
Maybe we need to put some filter so we are not sending a lot of CSP report errors to sentry.
What do you think?

Ataku-B:
@immortalcodes Hey Madhur, I've checked the code and QA'd it myself and it looks fine. I do think we should add some sort of filtering. Also, the new CSP should block the tfaforms on confidentiality-agreement.md, enquiry.md, and contact-us.md. Is this deliberate?

webapp/handlers.py
```diff
@@ -210,13 +220,219 @@ def _fetch_google_supported_domains():
...
+    "connect-src": [
+        "*.crazyegg.com",
+        "js.zi-scripts.com",
+        "px.ads.linkedin.com",
+        "ws.zoominfo.com",
+        "www.tfaforms.com",
+    ],
+    "style-src": ["www.tfaforms.com"],
+    "script-src": ["'unsafe-eval'"],
 }
```
immortalcodes, line 257: Hey @Skazitron 
I have put  *.crazyegg.com" in report mode, which means it will allow it but report it.
What was the decision on it, are we removing it or shall we keep it in report only mode?

immortalcodes:
@Ataku-B I have added de-duplication and filtering to not spam the sentry, could you take another look?

webapp/handlers.py
Skazitron: Let's put it into report mode.

Ataku-B (approved):
LGTM! Thank you for your efforts!

PR #2684, opened 2026-06-24
add llms-txt and llms-full.txt

Skazitron:
## Done
- Added `llms.txt` and `llms-full.txt`
- Added functions to dynamically populate these files from `/templates`
- Added an extensible `llms.yaml` to add custom entries to `llms.txt`

## QA
- Visit the [/llms.txt](https://canonical-com-2684.demos.haus/llms.txt) and [/llms-full.txt](https://canonical-com-2684.demos.haus/llms-full.txt)
- Ensure these are formatted as expected
- Ensure `robots.txt` is updated to allow the discovery of these two files.

## Issue / Card
https://warthogs.atlassian.net/browse/WD-36475


webapp/llms.py
```diff
@@ -0,0 +1,352 @@
...
+    for heading, links in _load_extra_sections():
+        lines.append(f"## {heading}")
+        lines.append("")
+        lines.extend(_link_bullet(link) for link in links)
+        lines.append("")
+
+    return "\n".join(lines).rstrip() + "\n"
+
+
+def generate_llms_full_txt(app):
```
SecondSkoll, line 294: [non-blocking] comment: This won't actually render a `llms-full.txt` file of everything on the website as documentation being served on the website are essentially separate websites served on specific paths - which means (AFAIK) they won't be attached to the existing nodes of the website.

This is probably fine, but it is something to be aware of.

llms.yaml
```diff
@@ -0,0 +1,30 @@
...
+#       links:
+#         - title: Link text
+#           url: https://example.com/page
+#           description: Optional one-line summary.   # may be omitted
+#
+# Links with no title or no url are skipped; a section with no valid links is
+# skipped. Editing this file and merging to main regenerates both files on the
+# next deploy.
+
+extra:
```
SecondSkoll, line 117: I assume this is already functional and tested, but the preview won't show this functionality working unless there's some content here. Would it be possible to add some so we can see it in the preview?

For example, just adding Juju would be:

```
- heading: Documentation
  links:
  - title: Juju
    url: https://canonical.com/juju/docs/juju-cli/page/llms.txt
    description: Juju is an open source orchestration engine for deploying, integrating, and managing applications across Kubernetes, VMs, and bare metal using software operators called charms.
```

Skazitron: @SecondSkoll of course! I'll add it shortly.

webapp/llms.py
Skazitron: @SecondSkoll, yep. This will only render stuff from the templates. Since the pro dashboard and other services are behind a reverse proxy, anything that we want to expose from them will need to be exposed within the extensible `llms.yaml` file

massigori:
As it stands the auto-generated section only sees the marketing templates, so our highest value,  agent-relevant content (the product docs ) can *only* reach the file through the manual `llms.yaml` append, sitting at the very bottom. That inverts the priority: we put our most LLM citable content is in the lowest position, most likely to be dropped. 

A few changes could fix that:

1. **Invert the order** Put curated yaml section first and push the auto-generated list into the bottom (ideally under an `## Optional` heading, which bots may skip under context pressure). Right now the append lands where a context-constrained bot is most likely to drop it

2. **Treat the root file as an index that links out to per-product `llms.txt` files** rather than trying to be comprehensive itself. That keeps it under the ~200K-token ingest budget and imho it is the optimal pattern for a site with many doc sub-sites behind a reverse proxy

3. **Confirm the generated entries carry real one-line descriptions, not just title + URL.** the description is the real value add compared to the `sitemap.xml`. if missing, agents get a link list with no signal/context about what to fetch

Minor but worth considering: the example `Juju` entry's description currently reads "Official Ubuntu Server documentation", probably a slip but it will happen. Worth a lightweight lint on `llms.yaml` (missing/duplicated descriptions) since these manual entries are exactly the ones that'll drift

llms.yaml
```diff
@@ -0,0 +1,122 @@
...
+#    describe Juju, not another product.
+#
+#  EXAMPLE — this one is live. Copy a link block (the four "- title" lines)
+#  to add more, and add another "- heading" block to start a new section.
+#
+extra:
+  - heading: Documentation
+    links:
+      - title: Juju
+        url: https://canonical.com/juju/docs/juju-cli/page
```
SecondSkoll:
```suggestion
        url: https://canonical.com/juju/docs/juju-cli/page/llms.txt
```

This would be how we connect this main llms.txt file to the ones docs generate.

SecondSkoll (commented):
If we're adding documentation via the extra field in the YAML - we're still adding it to the bottom of the file, rather than to somewhere near the top which was a concern Massi raised.

Otherwise this is perfectly functional for our needs.

tests/test_llms.py
```diff
@@ -0,0 +1,388 @@
...
+        self.assertEqual(overrides["/bad"], {})
+
+    def test_parses_valid_sections(self):
+        self._write_config(
+            """
+extra:
+  - heading: Documentation
+    links:
+      - title: Ubuntu Server docs
+        url: https://documentation.ubuntu.com/server/
```
SecondSkoll: This page is now just a redirect to ubuntu.com/server/docs

Skazitron:
@SecondSkoll Thanks for the speedy response. I'll push some code to ensure the custom links show up at the top instead of the bottom. I was also wondering if we should add support for an ordering mechanism to `llms.yaml` but that might be us overextending. Let me know what you think.

massigori (approved):
Nice work, approved from my side

SecondSkoll (approved):
Looks good to me. When merged I'll put some stuff together on the docs side and get all current docs entries added.

Ordering of the llms.txt file would probably be a nice to have - but I don't know how much we'd change about the current order. I don't see many teams adding custom entries.

.github/workflows/llms.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - "llms.yaml"
+
+jobs:
+  generate-llms:
+    if: github.repository == 'canonical/canonical.com'
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
muhammad-ali-pk: Valid comment.

.github/workflows/llms.yaml
```diff
@@ -0,0 +1,32 @@
+name: llms.txt generation for canonical.com
+on:
+  push:
+    branches:
+      - main
```
muhammad-ali-pk: is "on every push to main on templates" necessary? why not a cronjob that runs weekly or twice a week?

.github/workflows/llms.yaml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Regenerate llms.txt
+        run: |
+          curl --fail-with-body -X POST "https://canonical.com/llms.txt" \
+          -H "Authorization: Bearer ${{ secrets.SITEMAP_SECRET }}"
```
muhammad-ali-pk: Can we use a dedicated secret as `SITEMAP_SECRET` isn't directly related to this implementation.

package.json
```diff
@@ -9,7 +9,8 @@
     "build-js": "./scripts/build-modules.sh && node build.js",
     "format-python": "black --line-length 79 webapp tests",
     "format-prettier": "prettier -w 'static/js/*.{js,jsx,ts,tsx}' 'static/sass/*.scss'",
-    "lint-python": "flake8 --extend-ignore=E203 webapp tests && black --check --line-length 79 webapp tests",
+    "lint-python": "flake8 --extend-ignore=E203 webapp tests scripts && black --check --line-length 79 webapp tests scripts && yarn run lint-llms",
+    "lint-llms": "python3 scripts/lint_llms.py",
```
muhammad-ali-pk, line 13: Please add scripts to test llms and lint-llms

AGENTS.md
```diff
@@ -40,6 +41,7 @@
 - `.github/workflows/pr.yaml` - the most important CI definition; use it as the source of truth for lint/test/build expectations.
 - `.github/workflows/playwright.yaml`, `forms-test.yaml`, `percy-pr.yaml` - specialized browser/form/visual workflows.
 - `.github/workflows/sitemap.yaml` - on any `templates/**` push to `main`, POSTs to the live `https://canonical.com/sitemap_tree.xml` endpoint with `SITEMAP_SECRET` to regenerate the git-ignored `templates/sitemap_tree.xml` from the current `templates/` tree.
+- `.github/workflows/llms.yaml` - on any `templates/**` push to `main`, POSTs to the live `/llms.txt` and `/llms-full.txt` endpoints with `SITEMAP_SECRET` to regenerate those git-ignored files (same pattern as `sitemap.yaml`).
```
muhammad-ali-pk: Please also add information about `llms.yaml` file

llms.yaml
```diff
@@ -0,0 +1,122 @@
+# =============================================================================
```
muhammad-ali-pk, line 1: You'll also need to add this file in `rockcraft.yaml`

tests/test_lint_llms.py
```diff
@@ -0,0 +1,101 @@
+import unittest
```
muhammad-ali-pk, line 1: Please add a script for this test suite in package.json

tests/test_llms.py
```diff
@@ -0,0 +1,392 @@
+import logging
```
muhammad-ali-pk, line 1: Please add a script for this test suite in package.json

webapp/llms.py
```diff
@@ -0,0 +1,399 @@
...
+
+def _link_bullet(link):
+    """Render a curated link (with an explicit URL) as a bullet."""
+    line = f"- [{link['title']}]({link['url']})"
+    if link["description"]:
+        line += f": {link['description']}"
+    return line
+
+
+def _load_config():
```
muhammad-ali-pk, line 190: Load config is called in multiple places, resulting in parsing of yaml file multiple times. Maybe there's a way to parse this file only once, and accessed where needed?

webapp/llms.py
```diff
@@ -0,0 +1,399 @@
...
+def build_llms_view(app, key):
+    """Return a Flask view that serves (GET) or regenerates (POST) an
+    llms file, mirroring the sitemap_tree.xml generate-on-POST pattern.
+    """
+    config = LLMS_FILES[key]
+    file_path = os.path.join(os.getcwd(), config["filename"])
+
+    def generate_file():
+        content = config["generator"](app)
+        with open(file_path, "w") as f:
```
muhammad-ali-pk: maybe write to a temp file, and once done, replace with actual file, to prevent a half baked file being returned in the case of concurrent GET request.

muhammad-ali-pk (commented):
Great piece of work @Skazitron!
A few comments.

muhammad-ali-pk:
Getting `504 - timed out` on https://canonical-com-2684.demos.haus/llms-full.txt

.github/workflows/llms.yaml
Skazitron: We are rebuilding `llms.txt` either way each time the server starts/restarts. So this seems fitting.

llms.yaml
Skazitron: Good catch. Thanks for spotting.

.github/workflows/deploy.yaml
```diff
@@ -69,6 +69,15 @@ jobs:
...
           yarn run build
 
+      # Bake the LLM indexes into the image so every pod serves them from disk
+      # instead of rendering every page at request time (which times out for
+      # llms-full.txt). templates/ is primed by the rock, so the files ship.
+      - name: Pre-generate llms.txt and llms-full.txt
+        run: |
+          sudo apt-get update && sudo apt-get install --yes python3-setuptools
+          sudo pip3 install -r requirements.txt
+          python3 scripts/generate_llms.py
```
muhammad-ali-pk, line 79: You're trying to generate llms.txt and llms-full.txt files at build time.
But at this stage of workflow, relevant secrets (e.g., discourse env for discourse related pages) might not be available, therefore those pages will not render 200s and the generated files will not include their content.
What do you think?

scripts/generate_llms.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
```
muhammad-ali-pk, line 1: Logs from running this script locally: https://pastebin.canonical.com/p/k5c9pt8skR/

Seeing some failures

muhammad-ali-pk (commented):
Some comments from me, plus a few from copilot

PR #2712, opened 2026-07-03
feat: add sitewide json-ld

Skazitron:
## Done
- Added sitewide json-ld

## QA
- Run the branch locally
- Run `curl -s http://0.0.0.0:8002/ | grep -A30 'application/ld+json'`
- Check the `<head>` block for Organization.
- Confirm that this works behind proxied pages like account by running `curl -s http://0.0.0.0:8001/account | grep -c '"@type": "Organization"'`

## Issue / Card
https://warthogs.atlassian.net/browse/WD-36481

### DIFF
--- a/templates/base_index.html
+++ b/templates/base_index.html
@@ -91,7 +91,7 @@
                 "https://x.com/canonical",
                 "https://www.linkedin.com/company/canonical/",
                 "https://github.com/canonical",
-                "https://www.wikidata.org/wiki/Q1075063"
+                "https://www.wikidata.org/wiki/Q39064"
               ],
               "subOrganization": {
                 "@type": "Organization",
--- a/templates/base_index.html
+++ b/templates/base_index.html
@@ -74,6 +74,34 @@
 
           {% block extra_metatags %}{% endblock %}
 
+          <script type="application/ld+json" nonce="{{ csp_nonce }}">
+            {
+              "@context": "https://schema.org",
+              "@type": "Organization",
+              "@id": "https://canonical.com/#organization",
+              "name": "Canonical",
+              "url": "https://canonical.com/",
+              "logo": {
+                "@type": "ImageObject",
+                "url": "https://assets.ubuntu.com/v1/756d9941-canonical-logo.png",
+                "width": 364,
+                "height": 384
+              },
+              "sameAs": [
+                "https://x.com/canonical",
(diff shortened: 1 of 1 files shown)


templates/base_index.html
```diff
@@ -74,6 +74,34 @@
...
+                "@type": "ImageObject",
+                "url": "https://assets.ubuntu.com/v1/756d9941-canonical-logo.png",
+                "width": 364,
+                "height": 384
+              },
+              "sameAs": [
+                "https://x.com/canonical",
+                "https://www.linkedin.com/company/canonical/",
+                "https://github.com/canonical",
+                "https://www.wikidata.org/wiki/Q1075063"
```
Onibenjo:
```suggestion
                "https://www.wikidata.org/wiki/Q39064"
```

Onibenjo (changes requested):
a critical comment

Skazitron:
@Onibenjo Thanks for spotting that! Addressed.

templates/base_index.html
```diff
@@ -74,6 +74,34 @@
...
+                "https://www.linkedin.com/company/canonical/",
+                "https://github.com/canonical",
+                "https://www.wikidata.org/wiki/Q39064"
+              ],
+              "subOrganization": {
+                "@type": "Organization",
+                "@id": "https://ubuntu.com/#organization",
+                "name": "Ubuntu",
+                "url": "https://ubuntu.com/"
+              }
```
Onibenjo, line 101: is there something like a 

```
"brand": {
   "@type": "Brand",
   "name": "Ubuntu",
   "url": "https://ubuntu.com/"
 }
```

or was this specified by stakeholders?
Cause thinking of it, Ubuntu is a product/operating system, not an organization

Onibenjo (commented):
LGTM. A minor comment

templates/base_index.html
Onibenjo: I've seen Rhis comment on the other PR

Onibenjo (approved):

templates/base_index.html
rk-8: Fair point. The spec uses Organization deliberately because it enables the @id graph linking via subOrganization/parentOrganization, which only works between Organization types. I don't see it as a blocker issue.

rk-8 (commented):
THis looks good to me.

PR #2722, opened 2026-07-07
feat: adopt canonicalwebteam.discourse response cache and rate-limit handling

Onibenjo:
## Done

- Bump `canonicalwebteam.discourse` 7.1.1 → 7.6.0 and pass a `ResponseCache` to all 7 `DiscourseAPI` instances (charmhub, dqlite, MAAS docs + tutorials, engage pages, Mir docs, MicroK8s docs), so repeat renders serve from cache instead of calling Discourse on every request
- Handle the package's new `RateLimitedError`: Discourse rate limits now return a 503 with `Retry-After` (JSON body for JSON/API clients) instead of a 500

## Why

Crawler traffic recently exhausted the Discourse admin API quota (60 req/min, shared globally per instance) and 500'd Discourse-backed pages on ubuntu.com. **canonical.com's engage pages and Mir docs use the same API key against discourse.ubuntu.com**, uncached — so this site both suffers from and contributes to that shared-quota exhaustion. The caching + circuit breaker now live in the package (canonical/canonicalwebteam.discourse#227); this adopts them. Companion PR: canonical/ubuntu.com#16551.

## QA

- Check `/case-study`, `/events`, `/mir/docs`, `/maas/docs`, `/dqlite/docs`, `/microk8s/docs` load normally
- Repeat visits within the TTL (5–10 min) should not re-hit Discourse
- During a Discourse rate-limit, uncached pages return 503 + `Retry-After` rather than 500
- `pytest tests/test_app.py` passes; `yarn lint-python` clean

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ canonicalwebteam.blog==6.8.4
 canonicalwebteam.http==1.0.4
 canonicalwebteam.image-template==1.9.0
 canonicalwebteam.templatefinder==1.0.0
-canonicalwebteam.discourse==7.1.1
+canonicalwebteam.discourse==7.6.0
 canonicalwebteam.search==2.1.2
 canonicalwebteam.form-generator==2.2.0
 canonicalwebteam.directory-parser==1.2.11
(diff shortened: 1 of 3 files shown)


Onibenjo:
Not relevant anymore

PR #2799, opened 2026-07-21
feat: add manually authored llms.txt

Skazitron:
## Done
- Added manually authored `llms.txt`
- Added automatic generation of `llms-full.txt`
- Set cache TTL header for llms-full.txt to 24 hours.

## QA
- Visit /llms.txt on the demo.
  - Ensure the manually authored version under /templates shows up
  - Ensure content from `llms.yaml` is rendered as well.
- Visit /llms-full
- Ensure that the time taken to load up is not as long the 2nd time. (first time takes a while)

## Issue / Card
https://warthogs.atlassian.net/browse/WD-36475


templates/llms.txt
```diff
@@ -0,0 +1,180 @@
...
+- [MLflow](https://canonical.com/mlops/mlflow?format=md): MLflow, an open source platform for managing machine learning workflows.
+- [Canonical MLflow documentation](https://documentation.ubuntu.com/charmed-mlflow/): Official product documentation, explaining how to deploy and use Canonical MLflow
+- [Feast feature store](https://canonical.com/mlops/feast?format=md): Charmed Feast, a feature store for ML pipelines that integrates with Kubeflow and simplifies feature management.
+- [Canonical Feast documentation](https://documentation.ubuntu.com/charmed-feast/): Official product documentation, explaining how to deploy and use Canonical Feast
+
+## Application operations
+
+- [Juju](https://canonical.com/juju?format=md): An open source orchestration engine for software operators. Deploy, integrate, scale, and manage application lifecycles at any scale, on any infrastructure, with Juju and charms.
+- [Why Juju?](https://canonical.com/juju/why-juju?format=md): How Juju simplifies lifecycle management and operations.
+- [Juju product documentation](https://canonical.com/juju/docs/juju-cli/3.6/llms.txt): Official product documentation, explaining how to deploy and use Juju
```
Ataku-B, line 109: Since this link is an `llms.txt` link and not an `?format=md` link, I think it will always get rejected by `_is_renderable()` when building `llms-full.txt`

webapp/app.py
```diff
@@ -97,6 +98,11 @@
 with open("dynamic-sitemaps.yaml") as sitemaps_file:
     DYNAMIC_SITEMAPS = yaml.load(sitemaps_file.read(), Loader=yaml.FullLoader)
 
+# LLM-friendly site index (https://llmstxt.org/): templates/llms.txt (hand
+# written) plus curated extra links from llms.yaml. Built once at startup,
+# like the config above, rather than on every request.
+LLMS_TXT = llms.build_llms_txt("templates/llms.txt", "llms.yaml")
```
Ataku-B, line 104: Inside `build_llms_txt`, which calls `_load_extra_sections`, there is a safeguard against a missing llms.yaml file when opening it, but the same guard isn't present for `llms.txt` inside `build_llms_txt`

I tested removing both `llms.yaml` and `llms.txt` independently on my local environment, and the absence of the former is handled well, but the absence of the latter causes the build to fail. Ideally neither file should be removed recklessly in the future but it's still better to handle a potential build failure.

Ataku-B (dismissed):
👍 Great work! I QA'd this both locally and on the demo and it works well.

I left 2 minor comments, so I'm putting this as approved with changes.

webapp/app.py
Skazitron: Good catch. I think we should add a CI check for this to ensure we have proper formatting as well.

Skazitron:
@Ataku-B thank you for the comments. I've made a few changes:
- Added safeguards both in the backend and as a CI check.
- Added formatting check for llms.txt and llms.yaml
- Consolidated all of the llm python code into `webapp/llms.py`
- As of writing this comment, staging.canonical.com is running this branch, meaning it should deploy to prod without issues.

Please take another look.

.github/workflows/deploy.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - llms-txt-revamp
```
Skazitron: Needs to be removed before merging.

.github/workflows/deploy.yaml
```diff
@@ -147,55 +177,55 @@ jobs:
           juju refresh canonical-com-careers --path ./canonical-com_ubuntu-22.04-amd64.charm --resource flask-app-image=${{ needs.publish-image.outputs.image_url }}
           juju wait-for application canonical-com --query='name=="canonical-com" && (status=="active" || status=="idle")'
 
-  deploy-production:
```
Skazitron, line 150: Needs to be added uncommented before merging.

Ataku-B (approved):
LGTM 👍 
Thank you for your hard work!

Ataku-B (approved):
LGTM!
I think we're good to go

PR #2806, opened 2026-07-22
ci: cache packed charm and yarn deps to speed up deploys

Onibenjo:
## Done

Speed up the deploy pipeline by caching build inputs:

- **Cache the packed charm** (keyed on `charm/**`). The charm is a generic 12-factor flask-framework charm that only changes when `charm/` changes — which almost no deploy does. On a cache hit we skip LXD + `charmcraft pack` entirely. Measured: `pack-charm` **638s → ~6–11s**, taking it off the critical path.
- **Cache yarn deps** in `pack-rock` via `setup-node` (+ pin Node 20), so `yarn install` stops re-downloading every run. Minor gain — `pack-rock` is dominated by `rockcraft pack` — but harmless and better hygiene.

Net: the charm job is no longer the bottleneck; the pipeline floor is now the `pack-rock → publish-image → deploy-staging` chain. No behaviour change — the built charm and image are identical.

## QA

- Trigger a deploy that doesn't touch `charm/`: the "Restore cached charm" step reports a cache hit and the Setup LXD / Charmcraft / Pack charm steps are skipped.
  - Check https://github.com/canonical/canonical.com/actions/runs/29736297232
- After
  - https://github.com/canonical/canonical.com/actions/runs/29737088083

## Issue / Card

[WD-37854](https://warthogs.atlassian.net/browse/WD-37854)

## Screenshots
Before
image

After
image

[WD-37854]: https://warthogs.atlassian.net/browse/WD-37854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -26,6 +26,9 @@ concurrency:
 env:
   CHARMCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS: true
   ROCKCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS: true
+  # Pinned so the charm cache stays reproducible: edge drift can't leave a
+  # stale .charm cached under an unchanged charm/** key. Bump to invalidate.
+  CHARMCRAFT_REVISION: "8137"
 
 jobs:
   pack-charm:
@@ -45,7 +48,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ./charm/*.charm
-          key: charm-${{ hashFiles('charm/**') }}
+          key: charm-${{ runner.os }}-${{ runner.arch }}-cc${{ env.CHARMCRAFT_REVISION }}-${{ hashFiles('charm/**') }}
 
       - name: Setup LXD
         if: steps.charm-cache.outputs.cache-hit != 'true'
@@ -53,7 +56,7 @@ jobs:
 
       - name: Setup Charmcraft
         if: steps.charm-cache.outputs.cache-hit != 'true'
-        run: sudo snap install charmcraft --classic --channel=latest/edge
+        run: sudo snap install charmcraft --classic --revision ${{ env.CHARMCRAFT_REVISION }}
 
       - name: Pack charm
         if: steps.charm-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -62,11 +79,14 @@ jobs:
...
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'yarn'
 
       - name: Build Assets
         run: |
-          yarn install
+          yarn install --frozen-lockfile
```
muhammad-ali-pk, line 89: This can cause deploy failures if package.json and yarn.lock are not in sync. Since this is usually flagged in PR CI, we wouldn't want deploys to hard fail.
Subjective, lemme know what you think.

muhammad-ali-pk (approved):
Looks good to me. Though, since we need both `pack-charm` and `pack-rock` actions completed before moving on to next deployment steps, and `pack-rock` still takes around `7.5 minutes`, the deployment time still remains the same more or less.
Wondering if we can reduce `pack-rock` time somehow, to actually reduce deployment time.

Onibenjo:
> Wondering if we can reduce pack-rock time somehow, to actually reduce deployment time.

Yeah, we actually need the pack-rock to build and that can't be cached.
Only steps I can see is the LXD setup and all.


Atleast for now a 2-3min deploy time is gained and seeing we do sometimes have 5 deployments per day, we can see 15mins+ savings in build time

@muhammad-ali-pk

.github/workflows/deploy.yaml
Onibenjo: Turns out PR CI isn't actually catching this. pr.yaml runs `yarn install --immutable`, which yarn v1 silently ignores (checked locally, it exits 0 and happily rewrites the lockfile). So nothing enforces sync anywhere right now.

Given that, I'd keep the flag here. Failing at pack-rock is cheap since nothing has touched Juju yet, while a silent re-resolve means we ship dep versions CI never saw. 
I'll fix the PR workflows to use `--frozen-lockfile` in a follow-up, and then this check should almost never trip.

PR #2854, opened 2026-08-11
Add MAAS product pricing centrally

anthonydillon:
## Done

1. Move the pricing data for MAAS to the central `products.yaml` file and use that in multiple places on the webpage. 
2. Added `offers` as a field in the schema.org which is required for `Product` type objects

_Plan to do more products but one at a time to ensure through QA_

## QA

- Open the demo
- Go to /maas
- Open the production site and see they are the same


pedoch (approved):
LGTM! 

There a comment from copilot worth checking out.

PR #2875, opened 2026-08-20
fix: add lastmod entries to sitemaps

Skazitron:
## Done
- Script to auto-generate `lastmod` field generation.
- Fly-by:
  - Fixed lastmod not being generated in `sitemap_tree.xml`

## QA
- Deploy the branch to staging.
- Visit [partners/sitemap.xml]()
- Visti [/careers/sitemap.xml]()
- Ensure that the lastmod field is displayed with a date.

## Issue / Card
https://warthogs.atlassian.net/browse/WD-38046


scripts/generate-lastmod-manifest.py
```diff
@@ -0,0 +1,97 @@
...
+    for line in result.stdout.splitlines():
+        if line.startswith(RECORD_MARKER):
+            date = line[len(RECORD_MARKER) :]
+            continue
+
+        path = line.strip()
+        if not path.startswith(TEMPLATES_PATHSPEC + "/"):
+            # --name-only lists every file touched by the commit, not
+            # just ones under our pathspec; skip the rest.
+            continue
```
immortalcodes, line 56: Is it not already covered when in line 48 above when you mention the TEMPLATE_PATHSPEC in git process

scripts/generate-lastmod-manifest.py
```diff
@@ -0,0 +1,97 @@
...
+
+        rel_path = path[len(TEMPLATES_PATHSPEC) + 1 :]
+        # git log is newest-first, so the first commit we see touching
+        # a path is its most recent change.
+        manifest.setdefault(rel_path, date)
+
+    return manifest
```
immortalcodes, line 63: We are literally traversing the whole git history here for the repository which will become time-consuming as the commits grow.
What I would do is instead of building this whole file again at the build-time, I would keep the file in the repository, tag it with the date it was last built and then run the workflow to update it with commits that have taken place after that date-time

scripts/generate-lastmod-manifest.py
immortalcodes, line 1: I don't think `webapp/` is the best location for this file, we might want to move it to say `scripts/` for example, because this is essentially a build step and not part of flask app.

webapp/views.py
```diff
@@ -366,6 +368,57 @@ def add_cookie_to_url(match):
...
+def _load_lastmod_manifest():
+    """
+    Load templates/lastmod-manifest.json: a map of template file path
+    (relative to templates/) to the date it was last changed in git.
+
+    The production image doesn't ship .git (or the git history needed to
+    read it), so this manifest -- generated from git log by
+    scripts/generate-lastmod-manifest.py before the app is packaged -- is
+    how sitemap lastmod dates survive into the running app. Cached for
+    the life of the process: it's a static build artefact.
```
immortalcodes: While I understand the detailed comments are useful, we can trim these comments to just couple of lines or even format our code in a way that comments are not needed at all, bigger comments increase the file size and also consume more time to read.
The code should itself act a understanding of itself.

webapp/app.py
```diff
@@ -221,9 +225,37 @@ def index_sitemap():
...
+        "solutions/telco/5g-edge.html",
+    ),
+    (
+        "https://canonical.com/solutions/telco/5g-core",
+        "solutions/telco/5g-core.html",
+    ),
+    ("https://canonical.com/company", "company/index.html"),
+    ("https://canonical.com/knowledge", "knowledge/index.html"),
+]
+
```
immortalcodes: It would not be a good idea to populate these dicts in the app.py, may be we abstract the dicts into separate yaml?

webapp/app.py
```diff
@@ -393,10 +428,65 @@ def handle_careers_sitemap():
...
+        "https://canonical.com/careers/company-culture/diversity",
+        "careers/company-culture/diversity.html",
+        "monthly",
+    ),
+    (
+        "https://canonical.com/careers/company-culture/sustainability",
+        "careers/company-culture/sustainability.html",
+        "monthly",
+    ),
+]
```
immortalcodes: Also do we need to manually add these paths? As far as I remember, were we not using sitemap generator to dynamically generate paths?
cc: @britneywwc

webapp/app.py
```diff
@@ -812,9 +902,57 @@ def partner_details(partners_api):
...
+    pages = [
+        {
+            "url": url,
+            "last_modified": get_file_last_modified(TEMPLATES_DIR / path),
+        }
+        for url, path in PARTNERS_SITEMAP_PAGES
+    ]
+
+    xml_sitemap = flask.render_template("partners/sitemap.xml", pages=pages)
     response = flask.make_response(xml_sitemap)
```
immortalcodes: I suggest we should also abstract the sitemap generation functions in a separate file and then import and call them here, this makes the entire code more modular and clean.

webapp/views.py
```diff
@@ -515,6 +579,30 @@ def get_knowledge_sections():
...
+    templates_dir = Path(flask.current_app.root_path).parent / "templates"
+    index_file = templates_dir / "knowledge" / "index.html"
+
+    last_modified = get_file_last_modified(index_file)
+    if sections:
+        last_modified = max(
+            last_modified, *(s["last_modified"] for s in sections)
+        )
+
+    return last_modified
```
immortalcodes: We can actually generalize it by passing in the path and then it can basically do the last modified for any parent path.

.github/workflows/deploy.yaml
```diff
@@ -77,6 +77,13 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: ${{ github.event.inputs.branch || github.ref }}
+          # Full history, not the default shallow clone: the lastmod
+          # manifest step below needs to walk git log for every template
+          # file, and the rock image itself ships no .git at all.
+          fetch-depth: 0
+
+      - name: Pre-generate sitemap lastmod manifest
+        run: python3 webapp/lastmod_manifest.py generate-lastmod
```
immortalcodes: Were you able to test it on staging? I think you would have to make changes to rockcraft.yaml so that it actually copies the generated manifest inside the OCI-image.

immortalcodes (changes requested):
I have left some comments for the current approach.
I would also like to suggest an alternative approach where we save the manifest file for path-date date from git and instead of serving the sitemap files at the runtime, we generate them and pack them as static files inside the OCI-image.

Skazitron:
Hey @immortalcodes, thank you so much for the review. I've implemented all your suggested changes except for two things.  

1. I'll stick with the git history lookup for now instead of the incremental manifest since it appears to be quite fast (takes less than 1 second to run). 
2. The yaml thing--I'll wait for @britneywwc's answer on whether a dynamic sitemap generator can help us out with it. I would like to close this ticket before EOW, so I suggest we proceed with the implementation we have and add a maintenance ticket once we have more context.

You can take another look if you like.

scripts/generate-lastmod-manifest.py
Skazitron: This is actually quite performant and doesn't take too much time.

britneywwc:
@immortalcodes @Skazitron  Sitemap for the careers bubble is on https://canonical.com/careers/sitemap.xml, they are built dynamically. We should remove the hardcoded sitemaps as it will be redundant

Skazitron:
@britneywwc since my changes will be modifying the structure of the sitemap files and serving the sitemaps at those endpoints, it will be tricky to use the sitemap.xml both as a source of truth while generating them dynamically. That's why we were asking if there's some other source of truth we can use to generate the files instead.

Unless we have something like that, I propose we keep the hardcoded approach since the alternatives add unnecessary complexity.

britneywwc:
@Skazitron We should not be hardcoding the sitemaps if they are templates on the repository. There is a sitemap generator that will dynamically build static pages into sitemaps on canonical.com/sitemap_tree.xml. As good practice, we should always look into ways to reduce the need to hardcode stuff as this will increase maintenance overhead.

I see that /careers and /partners are both static page bubbles on this repository. Simply removing them on this [`dynamic-sitemap.yaml`](https://github.com/canonical/canonical.com/blob/main/dynamic-sitemaps.yaml#L3) file should be sufficient for the scope of the ticket. It will build the sitemap with the lastmod date taken from GitHub

.github/workflows/deploy.yaml
```diff
@@ -188,55 +189,55 @@ jobs:
...
-          juju wait-for application canonical-com --query='name=="canonical-com" && (status=="active" || status=="idle")'
+#   deploy-production:
+#     runs-on: [self-hosted, self-hosted-linux-amd64-noble-private-endpoint-medium]
+#     if: |
+#       always() &&
+#       needs.pack-charm.result == 'success' &&
+#       needs.publish-image.result == 'success' &&
+#       (needs.deploy-staging.result == 'success' || needs.deploy-staging.result == 'skipped') &&
+#       (
+#         github.event_name == 'push' ||
```
immortalcodes: Small reminder to uncommit them before merge

scripts/generate-lastmod-manifest.py
```diff
@@ -10,7 +10,7 @@
 webapp/views.py:get_file_last_modified, which reads it.
```
immortalcodes: I think we can improve this comment and strip out unnecessary details.

scripts/generate-lastmod-manifest.py
immortalcodes, line 53: Is this check not redundant as I pointed out in my last comment.
```
 [
            "git",
            "log",
            f"--format={RECORD_MARKER}%cs",
            "--name-only",
            "--",
            TEMPLATES_PATHSPEC,
        ],
```

immortalcodes (changes requested):

immortalcodes:
> Hey @immortalcodes, thank you so much for the review. I've implemented all your suggested changes except for two things.
> 
> 1. I'll stick with the git history lookup for now instead of the incremental manifest since it appears to be quite fast (takes less than 1 second to run).
> 2. The yaml thing--I'll wait for @britneywwc's answer on whether a dynamic sitemap generator can help us out with it. I would like to close this ticket before EOW, so I suggest we proceed with the implementation we have and add a maintenance ticket once we have more context.
> 
> You can take another look if you like.

Hey Shahir,
Thanks for working on the comments.
1. It is not much work implementing the incremental manifest, since you would need to just have a existing manifest + date when it was made and keep updating it on main. You can either save it as a GH artifact or even as a file in repo.
Are you sure it takes just 1 second to run the entire thing, even if that was the case we should aim for optimizations especially when it's not very difficult to implement.
2. I agree with @britneywwc 's comment regarding sitemaps.

scripts/generate-lastmod-manifest.py
Skazitron: Sorry, missed this last time. Done!

Skazitron:
@immortalcodes the current version (full rebuild) takes 0.16s to run locally and it only runs during the build step, so the performance gains would be very minimal. Moreover, committing the manifest to main would require CI steps and add more points of failure imo.

scripts/generate-lastmod-manifest.py
Skazitron: Not redundant, `--name-only` lists every file the commit touched, not just ones under templates/, so this filters those out.

Skazitron:
@immortalcodes implemented the changes suggested by @britneywwc. Would you take another look?

scripts/generate-lastmod-manifest.py
immortalcodes: What is the use of TEMPLATES_PATHSPEC here?

immortalcodes (approved):
LGTM! 
Could you please trim comments to be more precise and please remember to revert comments/changes in deploy workflow before merging

scripts/generate-lastmod-manifest.py
Skazitron: It scopes the git log to templates/ and strips that prefix to get manifest keys.

PR #2876, opened 2026-08-20
fix: add conditional meta tag for noindex when no articles are present

petesfrench:
## Done

- add conditional meta tag for noindex when no articles are present

## QA

- `curl -s http://localhost:8002/knowledge/company | grep -i 'name="robots"'` will return `<meta name="robots" content="noindex" />`
- `curl -s http://localhost:8002/knowledge/cloud-and-infrastructure | grep -i 'name="robots"'`  will return nothing

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-38391

### DIFF
--- a/templates/knowledge/_base_knowledge_category.html
+++ b/templates/knowledge/_base_knowledge_category.html
@@ -7,6 +7,12 @@
   is-paper
 {% endblock body_class %}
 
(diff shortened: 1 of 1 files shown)


britneywwc (approved):
Thanks, LGTM

PR #2902, opened 2026-08-31
fix: inject nonce into gtm container script

Skazitron:
## Done
- added nonce for gtm containers to abide by CSP.

## QA
- Navigate to the landing page
- Open browser console and ensure that there are no GTM errors relating to CSP.

## Issue / Card
https://warthogs.atlassian.net/browse/WD-38667

### DIFF
--- a/templates/base_index.html
+++ b/templates/base_index.html
@@ -240,6 +240,9 @@
                   j = d.createElement(s),
                   dl = l != 'dataLayer' ? '&l=' + l : '';
                   j.async = true;
+                  // nonce lets gtm.js propagate it to tags the container injects
+                  j.setAttribute('nonce', '{{ csp_nonce }}');
+                  j.nonce = '{{ csp_nonce }}';
                   j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl;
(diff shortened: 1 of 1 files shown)


immortalcodes:
https://github.com/canonical/canonical.com/blob/1160a4c51dd86b3cac6c08a7965747fc9a13ccee/webapp/handlers.py#L156
The CSP header seems to already have `ad.doubleclick.net`

Skazitron:
@immortalcodes please take a look at the comment from Gianluigi on the [parent ticket](https://warthogs.atlassian.net/browse/WD-38498)

immortalcodes (approved):
LGTM!
Just a comment, browser would hide nonce(look at script tag) but with the new code the nonce is clearly visible which could be utilized by inspected DOM by other actors.

Image

immortalcodes (approved):
LGTM!
Just a comment, browser would hide nonce(look at script tag) but with the new code the nonce is clearly visible which could be utilized by inspected DOM by other actors.

ImageLGTM

Skazitron:
The only way to deal with this would be to minify it, but that would require sitewide changes. We can discuss it during a Sites Dev Stuff meeting.

PR #2929, opened 2026-09-04
fix: add noindex to pages that should not be crawled

Skazitron:
## Done
- Added noindex to pages that should not be crawled.

## Issue / Card
https://warthogs.atlassian.net/browse/WD-38744


Ataku-B (approved):
LGTM

